PR #7, opened 2023-05-28
Create a C++ sample for AB-testing plugin.

amalykh:
Fixes #<issue_number_goes_here>

- [ ] Tests pass


samples/ab_testing/test.cc
```diff
@@ -0,0 +1,64 @@
...
+    EXPECT_EQ(res.http_code, 0);
+    std::string path = res.headers[":path"];
+    if (path == "example.com/file1.blah") {
+      EXPECT_THAT(res.headers, ElementsAre(Pair(":path", "example.com/file1.blah")));
+      file1_served = true;
+    } else {
+      EXPECT_THAT(res.headers, ElementsAre(Pair(":path", "example.com/file2.blah")));
+      file2_served = true;
+    }
+  }
```
leonm1, line 59: Nit: lines 41 and 46 are two test cases deserving their own TEST_P(HttpTest, UnrelatedPathUnchanged) and TEST_P(HttpTest, ExperimentPathExercised)

amalykh:
Hi [leonm1](https://github.com/leonm1), I have resolved the outstanding comments. Please take a look and see if we can merge the PR in.

samples/ab_testing/plugin.cc
```diff
@@ -0,0 +1,46 @@
...
+
+#include "proxy_wasm_intrinsics.h"
+
+class MyHttpContext : public Context {
+ public:
+  explicit MyHttpContext(uint32_t id, RootContext* root) : Context(id, root) {}
+
+  FilterHeadersStatus onRequestHeaders(uint32_t headers,
+                                       bool end_of_stream) override {
+    std::string path = getRequestHeader(":path")->toString();
```
martijneken, line 27: toString() forces a string copy, whereas WasmDataPtr->view() will return std::string_view to avoid copies.

You may also want an existence check on the :path header, although to be fair it should always exist.

samples/ab_testing/plugin.cc
```diff
@@ -0,0 +1,46 @@
...
+
+    const std::string default_file = "/file1.blah";
+    const std::string experiment_file = "/file2.blah";
+    if (path.length() >= default_file.length() &&
+        std::equal(default_file.rbegin(), default_file.rend(), path.rbegin())) {
+      bool serve_alternative_file = (rand() % 2 == 1);
```
martijneken, line 33: A little more challenging, but I wonder if we could demonstrate how to implement arbitrarily weighted choices, e.g. 80% A, 20% B.

samples/ab_testing/plugin.cc
```diff
@@ -0,0 +1,46 @@
...
+  explicit MyHttpContext(uint32_t id, RootContext* root) : Context(id, root) {}
+
+  FilterHeadersStatus onRequestHeaders(uint32_t headers,
+                                       bool end_of_stream) override {
+    std::string path = getRequestHeader(":path")->toString();
+
+    const std::string default_file = "/file1.blah";
+    const std::string experiment_file = "/file2.blah";
+    if (path.length() >= default_file.length() &&
+        std::equal(default_file.rbegin(), default_file.rend(), path.rbegin())) {
```
martijneken, line 32: I got a little lost in this logic -- is it a substring check on the full path? Worth a comment.,

You might also consider using URL parsing (and manipulation below) from boost, which I am adding in PR #17.

martijneken:
It's worth adding a link to this sample to the README (maybe once you have the Rust sample ready)

samples/ab_testing/plugin.cc
amalykh: Great point, I have added a check for the path (just in case) and changed this line to use string view.

samples/ab_testing/plugin.cc
amalykh: Agree, this sounds like a useful use-case to show. I have modified the code correspondingly.

samples/ab_testing/plugin.cc
amalykh: I have mode some refactoring here by moving this if condition to a separate function and adding around some comments. Please, take a look if it's more readable or whether you would like me to improve it further.

As for using the Boost library: I have a mild preference to have as few dependecies as we can get. I think it's more beneficial to focus on a specific thing that we are trying to showcase with this sample.

amalykh:
Hi [martijneken](https://github.com/martijneken), thanks for your review and sorry for not noticing that it has happened for a long time. I have addressed the comments and I would appreciate if you could make another pass.

As for the README file -- this should be straighforward, I'm happy to do it as part of the Rust PR.

martijneken (approved):
LGTM, thanks!

samples/ab_testing/plugin.cc
```diff
@@ -15,25 +15,45 @@
...
-    if (path.length() >= default_file.length() &&
-        std::equal(default_file.rbegin(), default_file.rend(), path.rbegin())) {
-      bool serve_alternative_file = (rand() % 2 == 1);
-      if (serve_alternative_file) {
-        const std::string truncated_path = path.substr(0, path.length() - default_file.length());
-        replaceRequestHeader(":path", truncated_path + experiment_file);
+    if (!getRequestHeader(":path")) {
+      return FilterHeadersStatus::Continue;
+    }
+    std::string_view path = getRequestHeader(":path")->view();
```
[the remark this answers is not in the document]
amalykh: Good point, done.

msampathkumar:
@amalykh - This PR seems ready, PTAL.

PR #12, opened 2023-06-23
docs: add region tags for add_header example

msampathkumar:
Purpose: To add region tags comments for the purpose of embedding code in Cloud Docs.

The new region tag is `serviceextensions_add_header`. This region tag is create from a string concatenation of repo name(`serviceextensions_`) and sample's folder name(`add_header`).

New region tags are added to the following files
* samples/add_header/plugin.cc#L15
* samples/add_header/plugin.rs#L15


leonm1 (approved):
The new tags LGTM, thanks!

PR #14, opened 2023-06-27
feat: update bazelrc flags so we can build with no args

leonm1:


.bazelrc
```diff
@@ -1,18 +1,15 @@
 # NOTE: this workspace Bazel config was copied from proxy-wasm-cpp-host.
 # Consider moving copts and linkopts into BUILD rules upstream.
 
-# Pass CC, CXX and PATH from the environment.
-build --action_env=CC
-build --action_env=CXX
-build --action_env=PATH
-
+# Service Extensions uses v8.
```
martijneken, line 4: Decision subject to change (and leaks implementation details). Maybe say "pick an engine to invoke sample plugins" or something similar.

.bazelrc
```diff
@@ -1,18 +1,15 @@
 # NOTE: this workspace Bazel config was copied from proxy-wasm-cpp-host.
```
martijneken, line 1: ... and edited?

martijneken (approved):

martijneken (changes requested):
Is this the right way to say "LGTM modulo comments" ?

martijneken (approved):

PR #16, opened 2023-06-29
feat: Introduce Cargo Bazel BUILD rules, add log and url to start

martijneken:
Uses cargo-raze to make adding dependencies fairly seamless.

- [x] Tests pass


cargo/BUILD.bazel
leonm1, line 1: Do we want to include cargo-raze genfiles as part of the repo?

The alternative is to not include them in the repo and add it as a step before you can build a rust plugin

The reasoning is that since the generated files can get out of date, we probably want to run cargo-raze on each fresh checkout _anyway_.

Another option would be to add a ci action that runs cargo-raze and ensures none of the generated files change? Or maybe if the files are out of date, the ci tests would fail anyway? Can you double check my thinking on this? 

```
bazel run @cargo-raze//:raze # or whatever
bazel test //...
```

Cargo.toml
```diff
@@ -0,0 +1,35 @@
...
+[profile.release]
+lto = true
+opt-level = 3
+codegen-units = 1
+panic = "abort"
+strip = "debuginfo"
+
+
+# To regenerate Cargo Bazel rules based on the above:
+#  $ sudo cargo install cargo-raze
```
leonm1, line 22: nit: if you change the instructions to use cargo-raze through bazel, don't forget to update this comment, too!

leonm1 (approved):

cargo/BUILD.bazel
martijneken: I think adding genfiles to the repo is the most hermetic way for now. The dependencies are for specific versions so they shouldn't get stale. Adding checkout/build handlers later SGTM.

README.md
```diff
@@ -67,6 +60,27 @@ will grow over time. The current feature set includes:
...
+(mainly HTTP headers and logging), but often in a simple way (behaviors may not
+match GCP exactly).
+
+## Rust and Cargo
+
+This project leverages cargo-raze to integrate Cargo with Bazel. In order to add
+new Rust library dependencies:
+
+*   Edit dependencies in Cargo.toml
+*   Install cargo-raze: `$ sudo cargo install cargo-raze`
```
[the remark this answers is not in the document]
martijneken: This is a great point. I'm having trouble adding a direct dependency due to incompatible versions of rules_rust between proxy-wasm-rust-sdk and cargo-raze. I'll keep poking at this, but ideally I'd have two separate workspaces.

Cargo.toml
martijneken: Done.

README.md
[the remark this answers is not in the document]
martijneken: Used a nested WORKSPACE. Gross but it definitely works.

PR #17, opened 2023-07-05
feat: Add sample for query string parsing + custom logging

martijneken:
Integrate Bazel with Boost via rules_boost. Some exception hackery required.

Branched from PR #16 

- [x] Tests pass


samples/query_log/plugin.cc
```diff
@@ -0,0 +1,45 @@
...
+
+#include "proxy_wasm_intrinsics.h"
+
+class MyHttpContext : public Context {
+ public:
+  explicit MyHttpContext(uint32_t id, RootContext* root) : Context(id, root) {}
+
+  FilterHeadersStatus onRequestHeaders(uint32_t headers,
+                                       bool end_of_stream) override {
+    auto path = getRequestHeader(":path");
```
leonm1, line 27: tiny nit: I assume we're following the Google c++ style guide, and per its guidance on `auto`, I think it'd be better to spell out the type; especially since this is a proxy_wasm type that users are likely to use.

https://google.github.io/styleguide/cppguide.html#Type_deduction

leonm1 (approved):
Looks great, glad you were able to figure out the boost_exceptions issue.

samples/query_log/plugin.cc
martijneken: Fair point, replaced here and below for boost types.

martijneken:
Adding @upeeth for comments/style review

PR #18, opened 2023-07-26
feat: Add sample using regex to rewrite a path

martijneken:
In C++ we use RE2 and Abseil.
In Rust we use the regex crate.

This example removes part of the path, but any URI mutation is feasible (path, query, fragment, etc).

The C++ sample is excluded from the build until we resolve Emscripten blockers:
https://github.com/proxy-wasm/proxy-wasm-cpp-sdk/pull/157
NOTE: leaving the code here because users can build the C++ plugin by using the PR above.

- [x] Tests pass


samples/regex_rewrite/plugin.rs
```diff
@@ -0,0 +1,71 @@
...
+// limitations under the License.
+
+// [START serviceextensions_regex_rewrite]
+use proxy_wasm::traits::*;
+use proxy_wasm::types::*;
+use regex::Regex;
+use std::rc::Rc;
+
+proxy_wasm::main! {{
+    proxy_wasm::set_log_level(LogLevel::Trace);
```
arman-rye, line 22: I think we should update these to Critical based on feedback that this sets the maximum log level vs previously thought minimum.

fengshi0730: Arman, sorry for the confusion. We were wrong about the rust log level setting. This is the correct way to set the minimum log level for rust as well: this means the min log level is Trace. @leonm1, FYI

martijneken: We could expand the noop_logs example to verify + demonstrate level-setting.

arman-rye: Ok thank you, I will make sure we communicate to customers accordingly.

leonm1:
Glad to see the tests in CI worked well!

samples/regex_rewrite/BUILD
```diff
@@ -0,0 +1,42 @@
...
+    deps = [
+        "//cargo:regex",
+        "@proxy_wasm_rust_sdk//:proxy_wasm",
+        "@proxy_wasm_rust_sdk//bazel/cargo:log",
+    ],
+)
+
+# Blocked on emscripten update to fix Abseil/RE2 compatibility:
+# https://github.com/proxy-wasm/proxy-wasm-cpp-sdk/pull/157
+#proxy_wasm_plugin_cpp(
```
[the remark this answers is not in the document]
martijneken: I went back and forth on this. I decided to leave it in because customers _can_ build the C++ plugin by using the linked ProxyWasmCppSdk PR. I added a note and confirmed this just now (and restored the absl version upgrade).

leonm1 (approved):

PR #23, opened 2023-08-30
fix: Make Rust work better in Bazel

martijneken:
* switch from cargo-raze to crate_universe
* target WASI platform
* optimize for execution speed
* add logging to stdout for unit tests
* [cleanup] set edition/crate-type to match rust_rules 

- [x] Tests pass


test/framework.cc
```diff
@@ -33,6 +33,11 @@ uint64_t TestContext::getMonotonicTimeNanoseconds() {
   unimplemented();
   return 0;
 }
+proxy_wasm::WasmResult TestContext::log(uint32_t log_level,
+                                        std::string_view message) {
+  std::cout << "LOG from testcontext: " << message << std::endl;
```
leonm1, line 38: Did you intend to leave this in here? Fine if so, just wanted to call attention in case it was not

leonm1 (approved):
Much cleaner than having a whole tree of cargo-raze generated files!

test/framework.cc
martijneken: Yep thanks, not really Rust specific but I found myself stashing and reinstating this over and over while working on plugins.

mpwarres:
(Sorry for late response) +1, looks like a great simplification!

PR #24, opened 2023-09-06
feat: include proxy-wasm using standard cargo tooling

leonm1:
This fixes the bug described in https://github.com/leonm1/service-extensions-samples/commit/9f97727b05103cbf6e7fd9013daa263ffe65a37c, where multiple log crates can be linked into the resulting wasm binary, as well as allowing standard rust IDE tooling to find all of the included crates.


Cargo.toml
```diff
@@ -35,5 +36,7 @@ targets = [
     "x86_64-apple-darwin",
     "x86_64-pc-windows-msvc",
     "x86_64-unknown-linux-gnu",
+    "wasm32-unknown-unknown",
+    "wasm32-wasi",
```
martijneken, line 40: You can delete all the other targets; this is what ProxyWasm supports amd the user guide recommends

Cargo.lock
```diff
@@ -0,0 +1,177 @@
+# This file is automatically @generated by Cargo.
```
martijneken, line 1: Dupe of cargo/Cargo.raze.lock?

leonm1: Done.

Cargo.toml
leonm1: Done.

PR #26, opened 2023-09-19
feat: Add benchmarking support to the samples repository

martijneken:
- rework .bazelrc: add benchmarking/testing flags, remove cruft
- add static registration macros for unit tests and benchmarks, so we automatically run them for all engines + plugin flavors
- add proxy_wasm_test macro to support tests + benchmarks in one file
- optimize the test framework for benchmarks (no logging, no tracing)
- write sample benchmarks for add_header and regex_rewrite
- update README with benchmarking command

I included a small unrelated change to neuter BoringSSL's RNG in tests.

- [x] Tests pass

### DIFF
--- a/test/framework.cc
+++ b/test/framework.cc
@@ -14,7 +14,7 @@
 
 #include "test/framework.h"
 
-#include <filesystem>
+#include <boost/dll/runtime_symbol_info.hpp>
 
 namespace service_extensions_samples {
 
@@ -156,7 +156,8 @@ std::string ReadDataFile(const std::string& path) {
 
 std::vector<std::string> FindPlugins() {
   std::vector<std::string> out;
-  for (const auto& entry : std::filesystem::recursive_directory_iterator(".")) {
+  for (const auto& entry : boost::filesystem::directory_iterator(
+           boost::dll::program_location().parent_path())) {
     if (entry.path().extension() == ".wasm") {
       out.push_back(entry.path().string());
     }
--- a/test/framework.cc
+++ b/test/framework.cc
@@ -37,7 +37,7 @@ uint64_t TestContext::getMonotonicTimeNanoseconds() {
 }
 proxy_wasm::WasmResult TestContext::log(uint32_t log_level,
                                         std::string_view message) {
-#ifdef NDEBUG
+#ifdef PROXY_WASM_TEST_SKIP_LOGS
   // No logging in release mode (for benchmarks).
   return proxy_wasm::WasmResult::Ok;
 #else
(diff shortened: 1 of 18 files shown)


martijneken:
Bleh. I turned off logging for opt builds, but our CI is testing a logging plugin with opt. The intent was to turn off logging for benchmarks only. I'll rework this logic to use a define.

martijneken:
Fixed. Ready for review.

test/framework.cc
```diff
@@ -137,14 +144,40 @@ TestHttpContext::Result TestHttpContext::SendResponseHeaders(
...
+  EXPECT_FALSE(file.fail()) << "failed to open: " << path;
+  std::stringstream file_string_stream;
+  file_string_stream << file.rdbuf();
+  return file_string_stream.str();
+}
+}  // namespace
+
+std::vector<std::string> FindPlugins() {
+  std::vector<std::string> out;
+  for (const auto& entry : std::filesystem::recursive_directory_iterator(".")) {
```
leonm1, line 159: tbh I'm surprised this works — why doesn't this recurse through all of bazel-bin, bazel-out, bazel-service-extensions-sampes, bazel-testlogs, as well as the other files in the repository?

leonm1 (approved):
LGTM modulo the concerns about `FindPlugins()`

I think it'd be a good idea to run our benchmarks in GitHub Actions to avoid bitrot. Would you like to include that in this PR, send a different PR, or have me take a look at adding it?

test/framework.cc
martijneken: Bazel is nice -- it organizes every binary into its own execution directory with all its data dependencies, and runs it from there (using symlinks to prevent file duplication). For example the add_header test runs from: .../execroot/__main__/bazel-out/k8-opt/bin/samples/add_header/plugin_test.runfiles/__main__/

But I see your point -- if I run the benchmark from bazel-bin/ directly then things break. 

Looks like I should search from the executable path, and there are various tools for this (whereami, boost::dll::program_location, etc). I'll see what works best.

martijneken: Used boost to scan for files next to the running binary. PTAL.

test/framework.cc
```diff
@@ -35,8 +37,13 @@ uint64_t TestContext::getMonotonicTimeNanoseconds() {
 }
 proxy_wasm::WasmResult TestContext::log(uint32_t log_level,
                                         std::string_view message) {
+#ifdef PROXY_WASM_TEST_SKIP_LOGS
```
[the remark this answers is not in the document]
leonm1: Ouch... that string concatenation will get expensive, fast. Great point.

test/framework.cc
leonm1: That seems like it will work well with data dependencies. LGTM.

leonm1 (approved):

PR #28, opened 2023-10-25
Add a 'plugins' namespace to sample tags

martijneken:
After the docs are updated to use these we can remove the old tags.

- [ ] Tests pass


martijneken:
Blocked on https://github.com/proxy-wasm/proxy-wasm-cpp-host/pull/365

PR #34, opened 2023-12-14
feat: Adding service callout example

jstraceski:


callouts/python/Dockerfile
```diff
@@ -0,0 +1,19 @@
...
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+FROM marketplace.gcr.io/google/debian11
+WORKDIR /home/callouts/python
+COPY . ./
+RUN apt-get update && apt-get install -y python3-grpc-tools
```
mkosieradzki, line 14: I would probably move this line before WORKDIR - if we want to have more generic layers lower in the stack

rshriram (approved):

PR #39, opened 2024-02-12
fix: treat header names as case insensitive

### LINKED ISSUE #9: Test framework uses case-sensitive header map implementation
Per https://github.com/GoogleCloudPlatform/service-extensions-samples/pull/8#discussion_r1210721293, our test framework should not use a case-sensitive header map implementation.

martijneken:
Also expand header mutation testing (include updates, removals).

Branched off #38. Fixes #9.

- [x] Tests pass

### DIFF
--- a/plugins/samples/add_header/tests.textpb
+++ b/plugins/samples/add_header/tests.textpb
@@ -1,14 +1,23 @@
 test {
   name: "AddsRequestHeader"
   request_headers {
-    result { has_header { key: "Message" value: "hello" } }
+    result {
+      has_header { key: "Message" value: "hello" }
+      has_header { key: "Welcome" value: "warm" }
+    }
   }
 }
 test {
-  name: "ExtendsRequestHeader"
+  name: "UpdatesRequestHeader"
   request_headers {
-    input { header { key: "Message" value: "hey" } }
-    result { has_header { key: "Message" value: "hey, hello" } }
+    input {
+      header { key: "Message" value: "hey" }
+      header { key: "Welcome" value: "cold" }
+    }
+    result {
+      has_header { key: "Message" value: "hey, hello" }
+      has_header { key: "Welcome" value: "warm" }
+    }
   }
 }
 test {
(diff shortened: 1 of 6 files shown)


leonm1:
Optional suggestion: since you have commit access to github.com/GoogleCloudPlatform/service-extensions-samples, you can create dependent pull requests that can be reviewed and merged separately.

To do so, you push `martijneken:configured-tests` directly to a branch under GoogleCloudPlatform/service-extensions-samples. Optionally, you can edit #38 to merge that branch into master, but it is not required. Then, you edit this pull request to merge `martijneken:ci-headers` into `martijneken:configured-tests`. That way, I can review this diff separately from #38 while that is not yet merged.

Once #38 is merged, you simply rebase this PR and change the merge target back to main.

If you'd prefer not to do that, I'll wait until #38 is merged to review this.

martijneken:
Yes I would like to learn this skill, but I need some hands on help. I'm having trouble just rebasing this branch onto the latest commit in configured-tests. Let's chat tomorrow.

martijneken:
Rebased! PTAL.

plugins/samples/add_header/tests.textpb
```diff
@@ -41,3 +58,38 @@ test {
     result { has_header { key: "Message" value: "foo, bar" } }
   }
 }
+# Tests for header name case insensitivity.
+test {
+  name: "CaseInsensitiveTest"
+  request_headers {
+    # Test framework itself doesn't distinguish case.
```
leonm1, line 65: nit: even after reading the code, this comment is confusing to me:

At first, I thought this comment was referring to the fact that the "assertion framework" / "config driven testing framework" ignored the case, but I now see that this is referring to the wasm host interface is what is ignoring the case.

But the wasm plugin itself can't ignore case, since if the wasm plugin enumerates all the headers, it will see case-sensitive strings, right?

But the test framework ignores the case set by the plugin in just the assertions and in get and set operations done by the plugin.

Could you take a whack at restating this comment?

leonm1 (approved):

plugins/samples/add_header/tests.textpb
martijneken: Yup, good call. Removed this test and added a clearer comment.

PR #44, opened 2024-02-16
fix: Refactoring the python callout example directory.

jstraceski:
Refactoring the python directory to inherit from the applicable existing python files.
Structured the docker examples in a way that allows us to have multiple separate examples without copying.
Removed old example server code.
Updated the README documentation to reflect the changes.
Updated the Dockerfile to not buffer the print logs from callout server.


callouts/python/extproc/example/http/http_service_callout.py
```diff
@@ -0,0 +1,107 @@
...
+from typing import Iterator
+
+import grpc
+from grpc import ServicerContext
+from grpc._server import _Server
+import service_pb2
+import service_pb2_grpc
+
+
+class HTTPCalloutService(BaseHTTPRequestHandler):
```
htuch, line 16: We don't support HTTP service callouts yet, why do we have this example?

rshriram (approved):

PR #41, opened 2024-02-16
feat: Adding new samples and restructuring folders structure

victorwoli:
Fixes #40 

## Use-cases

1. Insert/add HTTP request & response header
2. Update/Overwrite HTTP request & response header
3. Normalize a HTTP header on request
4. Read/insert/update HTTP request payload

PR #48, opened 2024-03-06
feat: Adding use-cases 1 to 4 in the new callout scructure

### LINKED ISSUE #40: Adding new samples and restructuring folders structure
## Use-cases

1. Insert/add HTTP request & response header
2. Update/Overwrite HTTP request & response header
3. Normalize a HTTP header on request
4. Read/insert/update HTTP request payload

## Details

UC1. We split the functionalities to only use the header manipulation.
UC2. New sample created to overwrite a particular header value from the client and from the origin with a new value.
UC3. New sample to check the host header and create a new HTTP header to shard requests based on device.
UC4. We use the code from the first sample and split it into a new package.

victorwoli:
Fixes #40 

Considering closed PR #41 comments

## Use-cases

1. Insert/add HTTP request & response header
2. Update/Overwrite HTTP request & response header
3. Normalize a HTTP header on request
4. Read/insert/update HTTP request payload


callouts/ssl_creds/localhost.crt
```diff
@@ -1,19 +0,0 @@
------BEGIN CERTIFICATE-----
```
rshriram, line 1: why are these files being deleted?

pweiber: the new structure is using the creds from python/extproc/ssl_creds, that's why we removed the external one

jstraceski: That additional ssl_creds was to be used in other callout examples, but I can add them later, I am working on a proto generation script currently.

callouts/python/extproc/service/callout_server.py
jstraceski, line 1: Is it possible that the formatting for this project was changed from 2 spaces to 4 spaces?
It's making it a bit difficult to see the changes, I am wondering if you could modify your formatting script?

callouts/python/extproc/tests/grpc_test.py
jstraceski, line 1: I am a bit confused, this path should be from the root directory `callouts/python/extproc` and not a relative path. Does this work when you run the test?

jstraceski (commented):
Mainly just formatting concerns, I can read through the changes but changing the format to the original 2 spaces would help with that.

callouts/python/extproc/service/callout_server.py
pweiber: Adjusted the formatting, let me know if it's fine now

callouts/__init__.py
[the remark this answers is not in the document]
pweiber: File deleted

callouts/python/__init__.py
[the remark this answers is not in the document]
pweiber: File deleted

callouts/python/extproc/tests/grpc_test.py
pweiber: We had an issue with the path and it was throwing a not found error, that's the reason we changed it but I found out what was the problem, should be good now for the tests. 
But for the service/callout_server.py I had to do a os.path.abspath workaround, I tried with the './extproc/ssl_creds/localhost.crt' but was not working when executing the examples, let me know if we can keep like that or use another strategy

jstraceski: Could you demonstrate how you were running the samples? Ideally you would run from the `callouts/python` directory and running with `python -m extproc.example.grpc.service_callout_example`. For me this allowed the old pathing. It would be ideal if we could use a direct path in the callout server settings as it would allow a user to provide a local link to a cert rather than relying on the project structure.

callouts/python/extproc/service/callout_server.py
```diff
@@ -61,23 +63,104 @@ def add_header_mutation(
...
+        append_action=3  # OVERWRITE_IF_EXISTS
+      )
+    )
+
+  if clear_route_cache:
+    header_mutation.response.clear_route_cache = True
+  return header_mutation
+
+
+def normalize_header_mutation(
```
jstraceski, line 103: There is no real conception of updating the headers from the callout in terms of what is represented in the proto.

would it be better to consolidate this update method into the add method, adding the ability to set the append action?

Something like this:

```
def add_header_mutation(
    add: list[tuple[str, str]] | None = None,
    remove: list[str] | None = None,
    clear_route_cache: bool = False,
    append_action: service_pb2.HeaderValueOption.HeaderAppendAction = None,
) -> service_pb2.HeadersResponse:
  """Generate a header response for incoming requests.

  Args:
    add: A list of tuples representing headers to add.
    remove: List of header strings to remove from the request.
    clear_route_cache: If true, will enable clear_route_cache on the response.

  Returns:
    The constructed header response object.
  """
  header_mutation = service_pb2.HeadersResponse()
  if add:
    for k, v in add:
      header_value_option = service_pb2.HeaderValueOption(
          header=service_pb2.HeaderValue(key=k, raw_value=bytes(v, 'utf-8'))
        )
      if append_action:
        header_value_option.append_action = append_action
      header_mutation.response.header_mutation.set_headers.append(header_value_option)
  if remove is not None:
    header_mutation.response.header_mutation.remove_headers.extend(remove)
  if clear_route_cache:
    header_mutation.response.clear_route_cache = True
  return header_mutation
```

callouts/python/extproc/service/callout_server.py
```diff
@@ -61,23 +63,104 @@ def add_header_mutation(
   return header_mutation
 
 
+def update_header_mutation(
```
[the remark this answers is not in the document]
jstraceski: https://github.com/GoogleCloudPlatform/service-extensions-samples/pull/48#discussion_r1520417097 would allow chaining of the add function as well.

callouts/python/extproc/tests/grpc_test.py
pweiber: Running like you mentioned worked, I found what was the problem, for some reason when I was running the example it was changing the path during execution, I was able to run with the old path and reverted the changes, should be good now.

callouts/python/extproc/service/callout_server.py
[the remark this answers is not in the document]
pweiber: Did the suggested changes and apparently it worked fine, I decided to keep the normalize_header_mutation function and call the add_header_mutation inside so we can keep a good separation, let me know if the changes are following what you expected.

jstraceski (approved):

PR #51, opened 2024-03-18
feat: Adding new samples for the python callouts

### LINKED ISSUE #49: Adding new samples for the python callouts pt2
## Use-cases
5. Generate custom HTTP responses on arbitrary field
6. Log a custom variable to cloud logging

## Details
UC5. Generate custom HTTP responses based on any of the readable parameters above; regardless of whether there is actual server behind it (e.g. mocking an external service, or returning information about the server pool, returning a response to a crossdomain.xml file, etc)

UC6. Perform a basic check on the request and authorize the request or reject the request - log the authorization decision to Cloud Logging

pweiber:
Fixes #49

## Use-cases
5. Generate custom HTTP responses on arbitrary field
6. Log a custom variable to cloud logging

### DIFF
--- a/callouts/python/extproc/example/add_custom_response/Dockerfile
+++ b/callouts/python/extproc/example/add_custom_response/Dockerfile
@@ -1,0 +1,21 @@
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#    http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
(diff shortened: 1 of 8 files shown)


pweiber:
@jstraceski I tried some alternatives to this but only with the --break i was able to build it, you know a workaround or the specific apt we can use so we can avoid it?

> RUN apt-get update && apt-get install -y python3-pip python3-grpc-tools
> RUN pip install google-cloud-logging --break-system-packages

jstraceski (approved):

PR #53, opened 2024-03-21
feat: Adding new sample for the python callouts

pweiber:
Adding the redirect sample as a standalone example

## Use-cases
7. Perform a HTTP redirect

## Details
UC7 - For a given URL - redirect to another URL

### DIFF
--- a/callouts/python/extproc/tests/redirect_test.py
+++ b/callouts/python/extproc/tests/redirect_test.py
@@ -1,0 +1,226 @@
+# Copyright 2024 Google LLC.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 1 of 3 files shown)


callouts/python/extproc/tests/redirect_test.py
```diff
@@ -0,0 +1,226 @@
...
+
+
+class CalloutServerTest(callout_server.CalloutServer):
+
+  def on_request_headers(
+      self, headers: HttpHeaders, context: ServicerContext
+  ) -> HeadersResponse:
+    """Custom processor on request headers."""
+    return callout_server.add_header_mutation(
+        add=[('host', 'service-extensions.com'), (':path', '/extensions')],
```
rshriram, line 48: dont you have to set the immediate response field for redirect?

pweiber: I was able to redirect without it, using just like this, I called /callouts for example and it redirected to the /extensions

rshriram: sure that is for the same host. But in that case, it is not a redirect. Its a rewrite. The browser's address bar does not change. A redirect requires sending HTTP 301

pweiber: Hmmm got it, I'll address the changes then

PR #56, opened 2024-03-27
feat: Adding new sample for the python callouts

pweiber:
Adding the redirect sample as a standalone example

## Use-cases
7. Perform a HTTP redirect

## Details
UC7 - For a given URL - redirect to another URL

New PR based on the new structure from #55 and requested changes in #53 to use ImmediateResponse
- Minor fix in the add_header to use the new callout_tools instead of callout_server


jstraceski (approved):
Looks good! Thank you for following the new structure I implemented.

PR #65, opened 2024-04-08
feat: Adding use-case 8 in the callout scructure #64

### LINKED ISSUE #64: Check client request for a bad token referencing a list in configuration data
## Use-case

1. Check client request for a bad token referencing a list in configuration data

## Details

UC8. Reference a list of bad tokens in plugin configuration data to check if a given request should be authorized or not. Log the authorization decision made by the plugin.

victorwoli:
Fixes #64 

## Use-case

1. Check client request for a bad token referencing a list in configuration data

### DIFF
--- a/callouts/python/extproc/example/validate_token/service_callout_example.py
+++ b/callouts/python/extproc/example/validate_token/service_callout_example.py
@@ -38,13 +38,7 @@ class CalloutServerExample(callout_server.CalloutServer):
 
   For request header callouts we check the content of the request and
   authorize the request or reject the request.
-  The content being checked is if the header has the header header-check.
-  The decision is logged to Cloud Logging.
-
-  For request body callouts we check the content of the request and
-  authorize the request or reject the request.
-  The content being checked is if the body has the body body-check.
-  The decision is logged to Cloud Logging.
+  The content being checked is if the header has a valid token.
   """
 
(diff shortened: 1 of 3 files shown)


callouts/python/extproc/example/validate_token/service_callout_example.py
```diff
@@ -0,0 +1,59 @@
...
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+from grpc import ServicerContext
+from envoy.service.ext_proc.v3 import external_processor_pb2 as service_pb2
+from extproc.service import callout_server
+from extproc.service import callout_tools
+
+
+def validation_header_token(request_headers):
```
rshriram, line 21: I think the goal of this use case is to showcase how the end user can validate JWT tokens as part of an extension. towards that goal, I suggest we use some python package that takes in JWKS keyset and extracts the JWT token from the "Authorization: bearer <...>" header and validates it. If the token is invalid, you deny the request. If its valid, then you add bunch of headers that provide the decoded JWT fields.

pweiber: @rshriram considering these comments let me know if the #70 covers that, if you agree we can consider this UC part of the jwt PR and close this current one

PR #72, opened 2024-04-11
ObservabilityServer sample that counts requests.

sudharsh:
Also fix arguments in add_body example


callouts/python/extproc/example/Dockerfile
```diff
@@ -48,7 +48,9 @@ COPY $copy_path ./
 EXPOSE 443
 EXPOSE 80
 # Un-secured debugging port. Can be removed once integrated into a final project.
-EXPOSE 8080 
+EXPOSE 8080
+# Un-secured counters
+EXPOSE 10000
```
rshriram, line 53: why dont you move this Dockerfile to the observability_server just like the way other examples have their own docker file?

rshriram: Se https://github.com/GoogleCloudPlatform/service-extensions-samples/blob/main/callouts/python/extproc/example/add_body/Dockerfile for example

rshriram (approved):

PR #79, opened 2024-05-09
fix: Address a few obvious issues

orloffm:
This doesn't change any behaviors. It clears a few obvious issues I've encountered while trying to use the repository.

1. Proper links to documents from README.md.
2. Code formatting in two README.md files.
3. Typos, e.g. "reciving".
4. Fixed the `add_body` example and added tests for it.

### DIFF
--- a/callouts/python/README.md
+++ b/callouts/python/README.md
@@ -2,64 +2,69 @@
 __Copyrights and Licences__
 
 Files using Copyright 2023 Google LLC & Apache License Version 2.0:
-* [service_callout.py](./extproc/service/callout_server.py) 
-* [callout_tools.py](./extproc/service/callout_tools.py) 
+
+* [service_callout.py](./extproc/service/callout_server.py)
+* [callout_tools.py](./extproc/service/callout_tools.py)
 
 # Requirements
 
 * Python 3.11+
 * [buf](https://buf.build/docs/introduction)
-* [requirements.txt](./requirements.txt) 
+* [requirements.txt](./requirements.txt)
 
-# Quick start 
+# Quick start
 
 The minimal operation of this Python-based ext_proc server requires the `grpcio` python package as well as the protobuf generator tool [buf](https://buf.build/docs/introduction).
 
-
 The prefered method of installation is through a virtual enviornment, `venv`. To set up the virtual enviornment run:
 
-```
+```shell
+cd service-extensions-samples/callouts/python
 python -m venv env
-source env/bin activate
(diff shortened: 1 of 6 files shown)


callouts/python/README.md
```diff
@@ -2,64 +2,69 @@
...
 The minimal operation of this Python-based ext_proc server requires the `grpcio` python package as well as the protobuf generator tool [buf](https://buf.build/docs/introduction).
 
-
 The prefered method of installation is through a virtual enviornment, `venv`. To set up the virtual enviornment run:
 
-```
+```shell
+cd service-extensions-samples/callouts/python
 python -m venv env
-source env/bin activate
```
orloffm, line 23: Didn't work.

callouts/python/README.md
```diff
@@ -2,64 +2,69 @@
...
 pip install -r requirements-test.txt
 ```
 
 `buf` can be installed from [here](https://buf.build/docs/installation).
 
-The proto library files are generated with `buf` using: 
+The proto library files are generated with `buf` using:
 
-```
-buf -v generate https://github.com/envoyproxy/envoy.git#subdir=api --path envoy/service/ext_proc/v3/external_processor.proto --include-imports
```
orloffm, line 43: I've split lines like these.

callouts/python/extproc/example/add_body/service_callout_example.py
```diff
@@ -31,13 +32,13 @@ def on_request_body(
       self, body: service_pb2.HttpBody, context: ServicerContext
   ) -> service_pb2.BodyResponse:
     """Custom processor on the request body."""
-    return callout_server.add_body_mutation(body='-added-body')
+    return callout_tools.add_body_mutation(body='-added-body')
```
rshriram, line 35: why is this change being done?

rshriram (approved):

callouts/python/extproc/example/add_body/service_callout_example.py
jstraceski: It looks like the swap over to callout tools was never completed for this file, this is good, there should have been a test to catch this. Thank you for adding one.

callouts/python/README.md
jstraceski: Good catch.

callouts/python/README.md
jstraceski: Looks good to me.

jstraceski (approved):

PR #80, opened 2024-05-09
fix: Assorted fixes and tweaks to improve legibility and functionality.

jstraceski:
Fixing add_body_mutation function.
Updated examples to reflect.
Updated tests to reflect.
Modified example to actually add a body rather than replacing.
Updated test to reflect.
Added test cases for body_clear.
Modifying wording of callout_tools.py methods to better reflect the intended purpose.
Moving some example specific code out of the general callout_tools.py file. 
Modifying the normalization test to exercise all code paths from the example.
Modifying "request" to "callout" where applicable within callout_server.py.
Removing duplicated boilerplate class documentation from examples.
Cleaned up some examples, adding type hinting.
 
### Phase 2
Cleaning up README.md and adding development buf resources.
Modifying requirement files to not overlap and to be clearer.
Refactoring / cleaning up requirements.
Clarifying and updating Docker instructions.
Reworking deny tool, adding common functions to parse headers and body.
Renaming address method to denote that it is for internal use.
Updating README.md to reflect address name change.
Cleaning up custom response example and test.
Updating basic test to reflect address naming change.
Removing duplicate/unused Dockerfiles.
Reworking jwt example for Dockerfile README.md section.
Reworking cloud_log example to use custom Dockerfile.
Adding default logging to all examples.
Updating README.md with a possible linting edgecase.

- [x] Tests pass


callouts/python/extproc/example/basic_callout_server.py
```diff
@@ -57,7 +57,7 @@ def on_request_body(self, body: HttpBody, _) -> BodyResponse:
   def on_response_body(self, body: HttpBody, _) -> BodyResponse:
     """Custom processor on the response body."""
     logging.debug("Received response body callout: %s", body)
-    return add_body_mutation(body='new-body', clear_body=True)
+    return add_body_mutation(clear_body=True)
```
rshriram, line 60: why does the basic callout server need to touch body or clear the response?

callouts/python/extproc/service/callout_tools.py
```diff
@@ -140,23 +141,25 @@ def add_body_mutation(
...
-Args:
-  body: Text of the body.
-  clear_body: If set to true, the modification will clear the previous body,
-    if left false, the text will be appended to the end of the previous
-    body.
-  clear_route_cache: If true, will enable clear_route_cache on the response.
+  Args:
+    body: Body text to replace the current body of the incomming request.
+    clear_body: If true, will clear the body of the incomming request. 
+    clear_route_cache: If true, will enable clear_route_cache on the response.
```
rshriram, line 151: why do we have clear_route_cache for responses? thats useless isnt it?

callouts/python/extproc/example/basic_callout_server.py
jstraceski: This is intended just as an example server with an assortment of possible interactions. I can reduce the complexity to make it more of a basic example.

callouts/python/extproc/service/callout_tools.py
jstraceski: If we do not clear the route cache on a request body no-op response unless the envoy is configured to clear the route cache every time, A no-op request callout will cause the response callout to not be processed in every case. This is mainly for the request body as that is the one that is being processed before the response. It would be impractical to make one method for request bodies and one for response bodies.

rshriram: @tyxia is this a bug?

tyxia: This is not the expected behavior.
First, clear_route_cache is only applicable to mainstream **request** path. Second, I don't follow the logic why clear_route_cache on request path could affect response path behavior

@jstraceski I think this should require more debugging

jstraceski: So, the reason why the function says `clear_route_cache on the response` is because the returning datatype of this function is a `BodyResponse`, I have made a quick patch to make this more evident. This is the first pass I have done just on the callout_tools.py file, and some associated changes.

jstraceski:
I checked the SDK build step locally, it appears that the command is malformed somehow, is it possible that the command line arguments are not being read correctly?

jstraceski:
https://github.com/bufbuild/buf/issues/3000 looks like the buf folks may have some work to do.

rshriram (approved):

bufdev:
This should be good to go now with `buf` `v1.32.1` - apologies for the hassle. Let us know if you have any questions!

jstraceski:
> This should be good to go now with `buf` `v1.32.1` - apologies for the hassle. Let us know if you have any questions!

Thanks for the quick fix!

PR #82, opened 2024-05-17
feat: BODY Processing for plugins for benchmarking purposes.

dallen68:
feat:BODY Processing for plugins for benchmarking purposes.

additional fix: TestHttpContext is intended to live for a single request/response. We had an issue where multiple benchmark iterations were using the same context and this led to unintended behaviors.
e.g: When a plugin had logic to exit immediately after completing a task, in subsequent iterations the plugin would exit as if the task had completed that iteration.

- [ ] Tests pass


plugins/test/framework.cc
```diff
@@ -45,7 +85,19 @@ proxy_wasm::WasmResult TestContext::log(uint32_t log_level,
...
   return proxy_wasm::WasmResult::Ok;
 }
-
+proxy_wasm::BufferInterface* TestHttpContext::getBuffer(
+    proxy_wasm::WasmBufferType type) {
+  switch (type) {
+    case proxy_wasm::WasmBufferType::PluginConfiguration:
+      return TestContext::getBuffer(type);
+    case proxy_wasm::WasmBufferType::HttpRequestBody:
+      return &body_buffer_;
```
mpwarres, line 94: nit: this should return nullptr if the body buffer type doesn't match the current callback, e.g. if the current callback is onRequestBody but getBuffer() is called with type=HttpResponseBody.

plugins/test/framework.h
```diff
@@ -28,6 +28,42 @@
...
+  proxy_wasm::WasmResult copyFrom(size_t start, size_t length,
+                                  std::string_view data) override;
+
+  // proxy_wasm::BufferBase
+  void clear() override {
+    proxy_wasm::BufferBase::clear();
+    string_buffer_ = nullptr;
+  }
+  Buffer* set(std::string_view data) {
+    return static_cast<Buffer*>(proxy_wasm::BufferBase::set(data));
```
mpwarres, line 49: optional: if you `return this` then a cast isn't needed

plugins/test/framework.h
```diff
@@ -28,6 +28,42 @@
...
+  }
+  Buffer* set(std::unique_ptr<char[]> owned_data, uint32_t owned_data_size) {
+    return static_cast<Buffer*>(
```
mpwarres, line 52: optional: if you `return this` then a cast isn't needed

plugins/test/framework.h
```diff
@@ -28,6 +28,42 @@
...
+
+  Buffer* set(std::string* string) {
+    clear();
+    string_buffer_ = string;
+    return this;
+  }
+  std::string get() { return *string_buffer_; }
+
+ private:
+  std::string* string_buffer_;
```
mpwarres, line 64: Is this string owned by Buffer or not? Could the existing `data_` or `owned_data_` fields be used instead?

mpwarres (commented):
Looks good overall, just minor comments. Thanks!

plugins/test/framework.cc
```diff
@@ -18,13 +18,53 @@
...
+}
+
+proxy_wasm::WasmResult Buffer::copyFrom(size_t start, size_t length,
+                                        std::string_view data) {
+  if (string_buffer_) {
+    if (start == 0) {
+      if (length != 0) {
+        string_buffer_->erase(0, length);
+      }
+      string_buffer_->insert(start, data);
```
martijneken, line 48: AIUI this could be quite an expensive operation on std::string since it could shift/copy all of the std::string contents? I think we should compose a series of string_views here. When wasm publishes a new string, we take local ownership, but we leave alone any bits of the body that aren't being touched.

As a (complex) example, Envoy's Buffer implementation stores a collection of slices to handle data inserts at the front and back: https://github.com/envoyproxy/envoy/blob/ad0a10013c977a0f6a1e9de876d0a9eb22f1a187/source/common/buffer/buffer_impl.h#L773

We don't have to be quite as optimized, but we should avoid copying any content that the wasm doesn't mutate. My other comment about reducing string copies implies that ideally the Buffer should hold a mix of externally owned strings (the input) and internally owned mutations (received via setBuffer).

plugins/test/framework.cc
dallen68: I don't think we store that information in the TestContext.

plugins/test/framework.h
```diff
@@ -28,10 +29,51 @@
...
+
+  // proxy_wasm::BufferBase
+  void clear() override {
+    proxy_wasm::BufferBase::clear();
+    owned_string_buffer_ = "";
+  }
+  Buffer* set(std::string_view data) {
+    proxy_wasm::BufferBase::set(data);
+    return this;
+  }
```
mpwarres, line 61: I don't see the need to override the implementations of these methods that BufferBase provides.

The redefinition of set() also opens the door to possible problems: since BufferBase does not declare set() as virtual (which it probably should), the declaration of set() here doesn't get the dynamic dispatch that you would normally expect with a virtual method. IOW, if you invoke buffer->set(...) where buffer is a Buffer*, Buffer::set() will be called, but if buffer is a BufferBase*, BufferBase::set() will be called.

plugins/test/framework.h
```diff
@@ -28,10 +29,51 @@
...
+    if (made_body_mutation_) {
+      return owned_string_buffer_;
+    }
+    return std::string(data_);
+  }
+
+ private:
+  bool made_body_mutation_;
+  // Buffer for a chunk post-mutation.
+  std::string owned_string_buffer_;
```
mpwarres, line 73: I don't see the need to define these additional fields. The fields already inherited from BufferBase are:

```
  std::string_view data_;
  std::unique_ptr<char[]> owned_data_;
  uint32_t owned_data_size_;
```

With these, you can represent made_body_mutation_ by whether or not owned_data_ is set (i.e. non-null). So get() would be something like:

```
std::string get() {
  return owned_data_ != nullptr ? std::string(owned_data_.get(), owned_data_size_) : std::string(data_);
}
```

However, it's probably better to instead want to provide a string_view accessor, so that inspecting buffer contents doesn't require a copy:

```
std::string_view get() {
  return owned_data_ != nullptr ? std::string_view(owned_data_.get(), owned_data_size_) : data_;
}
```

plugins/test/framework.cc
```diff
@@ -130,21 +192,43 @@ TestHttpContext::Result TestHttpContext::SendRequestHeaders(
...
 }
+
+TestHttpContext::Result TestHttpContext::SendRequestBody(
+    const std::string_view body) {
+  phase_logs_.clear();
+  result_ = Result{};
+  body_buffer_.set(body);
+  result_.body_status =
+      onRequestBody(body_buffer_.size(), /*end_of_stream=*/false);
+  result_.body = body_buffer_.get();
```
mpwarres, line 208: This incurs a copy even if the body hasn't been changed.

To avoid a copy, I think you may need to change Result::body to be of type string_view, and also change Buffer::get() to return string_view as suggested in another comment.

plugins/test/framework.cc
mpwarres: I'd propose that for this stage, we start with a basic copy-on-write Buffer impl (which BufferBase provides 90% of), then change to something Cord-like as a further refinement later on. Just to get something simple and "good enough" landed for now.

plugins/test/dynamic_test.cc
```diff
@@ -178,7 +178,12 @@ void DynamicTest::TestBody() {
     CheckPhaseResults("request_headers", invoke.result(), stream, res);
   }
   if (cfg_.request_body_size() > 0) {
-    ADD_FAILURE() << "BODY processing not implemented yet";
+    for (auto& invoke : *cfg_.mutable_request_body()) {
+      auto res = stream.SendRequestBody(
+          std::move(*invoke.mutable_input()->release_content()));
```
mpwarres, line 183: I think this is a slight memory leak: while it moves the content of the string, the std::string object itself is never freed. To explain this, suppose I expand the code to the following equivalent snippet (just giving the pointer a name):

```
std::string* body_str = *invoke.mutable_input()->release_content();
auto res = stream.SendRequestBody(std::move(*body_str));
```

While the payload of the data that `body_str` contains is moved out of `body_str`, `body_str` itself remains, and nothing frees it.

The simplest way to ensure that the std::string* is properly deleted is to wrap it in a unique_ptr:

```
std::unique_ptr<std::string> body_str = absl::WrapUnique(*invoke.mutable_input()->release_content());
auto res = stream.SendRequestBody(std::move(*body_str));
```

This way, the payload of `body_str` is moved, and unique_ptr ensures that `body_str` itself is deleted.

plugins/test/dynamic_test.cc
```diff
@@ -189,7 +194,12 @@ void DynamicTest::TestBody() {
     CheckPhaseResults("response_headers", invoke.result(), stream, res);
   }
   if (cfg_.response_body_size() > 0) {
-    ADD_FAILURE() << "BODY processing not implemented yet";
+    for (auto& invoke : *cfg_.mutable_response_body()) {
+      auto res = stream.SendResponseBody(
+          std::move(*invoke.mutable_input()->release_content()));
```
mpwarres, line 199: Same comment as above re: memory leak.

plugins/test/dynamic_test.cc
```diff
@@ -280,19 +290,35 @@ void DynamicTest::BenchHttpHandlers(benchmark::State& state) {
...
+  std::vector<std::string> request_body_chunks;
+  for (const auto& request_body : cfg_.request_body()) {
+    request_body_chunks.emplace_back(request_body.input().content());
+  }
+  std::vector<std::string> response_body_chunks;
+  for (const auto& response_body : cfg_.response_body()) {
+    response_body_chunks.emplace_back(response_body.input().content());
+  }
   for (auto _ : state) {
     if (request_headers) {
```
mpwarres, line 302: Don't you need to move `auto stream = TestHttpContext(handle);` inside the loop?

plugins/test/dynamic_test.cc
```diff
@@ -280,19 +290,35 @@ void DynamicTest::BenchHttpHandlers(benchmark::State& state) {
...
       auto res = stream.SendRequestHeaders(*request_headers);
       benchmark::DoNotOptimize(res);
       BM_RETURN_IF_FAILED(handle);
     }
-    // Future: send request BODY here.
+    for (std::string& body : request_body_chunks) {
+      auto res = stream.SendRequestBody(body);
```
mpwarres, line 308: Note that `SendRequestBody(body)` copies `body`. This copy is unavoidable, since there's only one set of `request_body_chunks` vs. many iterations of the outer `for (auto _ : state)` loop, so can't `std::move` out of `request_body_chunks`.

However, if you want to make sure that the string copy is not included in benchmark numbers, you could explicitly copy the request and response body strings at the top of the loop, inside of a `PauseTiming()` + `ResumeTiming()` block (which you might need to add anyways around creation of the `TestHttpContext`, per earlier comment). IOW, you could do:

```
for (auto _ : state) {
  state.PauseTiming();
  auto stream = TestHttpContext(handle);
  std::vector<std::string> request_body_chunk_copies = request_body_chunks;
  std::vector<std::string> response_body_chunk_copies = response_body_chunks;
  state.ResumeTiming();

  // ...

  for (std::string& body : response_body_chunk_copies) {
    auto res = stream.SendResponseBody(std::move(body));
    benchmark::DoNotOptimize(res);
    BM_RETURN_IF_FAILED(handle);
  }
  // ...
}
  }
}
```

plugins/test/framework.cc
```diff
@@ -20,13 +20,54 @@
...
+  if (owned_string_buffer_) {
+    return owned_string_buffer_->length();
+  }
+  return proxy_wasm::BufferBase::size();
+}
+proxy_wasm::WasmResult Buffer::copyTo(proxy_wasm::WasmBase* wasm, size_t start,
+                                      size_t length, uint64_t ptr_ptr,
+                                      uint64_t size_ptr) const {
+  if (owned_string_buffer_) {
+    auto buffer_section = owned_string_buffer_->substr(start, length);
```
mpwarres, line 33: `substr()` returns a `std::string`, which means that this line incurs an extra copy in addition to the later (necessary) copy into the Wasm VM. To avoid this, do:

```
if (start + length > owned_string_buffer_.size()) {
  return WasmResult::InvalidMemoryAccess;
}
std::string_view s(owned_string_buffer_.data() + start, length);
if (!wasm->copyToPointerSize(s, ptr_ptr, size_ptr)) { ... }
```

plugins/test/framework.cc
```diff
@@ -20,13 +20,54 @@
...
+  }
+  return proxy_wasm::BufferBase::copyTo(wasm, start, length, ptr_ptr, size_ptr);
+}
+
+proxy_wasm::WasmResult Buffer::copyFrom(size_t start, size_t length,
+                                        std::string_view data) {
+  if (!owned_string_buffer_) {
+    owned_string_buffer_ = std::string(data_);
+  }
+  if (start == 0) {
```
mpwarres, line 48: I think this can be simplified significantly if you use `std::string::replace()`, which does pretty much everything you need. i.e. something like:

```
owned_string_buffer_.replace(start, length, data);
```

plugins/test/framework.cc
```diff
@@ -55,6 +96,26 @@ ContextOptions& TestContext::options() const {
...
+      break;
+    case proxy_wasm::WasmBufferType::HttpResponseBody:
+      if (current_callback_ == TestHttpContext::CallbackType::ResponseBody) {
+        return &body_buffer_;
+      }
+      break;
+    default:
+      return nullptr;
+  }
+  return nullptr;
```
mpwarres, line 117: nit: since the switch statement is exhaustive, this line can't be reached, so you should be able to remove the final `return nullptr;` that is outside of the switch statement.

plugins/test/framework.cc
```diff
@@ -134,22 +195,47 @@ TestHttpContext::Result TestHttpContext::SendRequestHeaders(
...
+  result_.header_status =
       onRequestHeaders(result_.headers.size(), /*end_of_stream=*/false);
   phase_ = proxy_wasm::WasmHeaderMapType(-1);  // ideally 0 would mean unset
   return std::move(result_);
 }
+
+TestHttpContext::Result TestHttpContext::SendRequestBody(std::string body) {
+  phase_logs_.clear();
+  result_ = Result{};
+  body_buffer_.set(std::move(body));
```
mpwarres, line 208: IIUC this is calling `BufferBase::set(std::string_view data)`, which won't take ownership. I think you need to define a new `Buffer::SetOwned(std::string data)` method:

```
Buffer* SetOwned(std::string data) {
  clear();
  owned_string_buffer_ = std::move(data);
  return this;
}
```

I'd avoid naming this new method `set()` because then it will be unclear whether code is invoking `Buffer::set(std::string)` or `BufferBase::set(std::string_view)`.

plugins/test/framework.cc
```diff
@@ -134,22 +195,47 @@ TestHttpContext::Result TestHttpContext::SendRequestHeaders(
...
+  result_.header_status =
       onResponseHeaders(result_.headers.size(), /*end_of_stream=*/false);
   phase_ = proxy_wasm::WasmHeaderMapType(-1);  // ideally 0 would mean unset
   return std::move(result_);
 }
 
+TestHttpContext::Result TestHttpContext::SendResponseBody(std::string body) {
+  phase_logs_.clear();
+  result_ = Result{};
+  body_buffer_.set(std::move(body));
```
mpwarres, line 231: Same as earlier comment, this should call a new `SetOwned()` method.

plugins/test/dynamic_test.cc
dallen68: https://github.com/GoogleCloudPlatform/service-extensions-samples/pull/95

plugins/test/framework.cc
dallen68: AIUI if you were to break out of that switch statement then the default case would not be called, so we would need to return outside of the switch.

plugins/test/dynamic_test.cc
dallen68: After looking at your other comment, I think it would be easier to include this fix in this PR.

plugins/test/framework.cc
dallen68: We don't want ownership at this point since there have been no body mutations. This lets us avoid unnecessary string copies in the case where there is no change.

plugins/test/framework.cc
dallen68: Same comment as earlier

mpwarres (approved):
LGTM

PR #86, opened 2024-05-29
feat: Go callout server

pweiber:
Adding a go callout server SDK

### Current examples

add_header_mutation
add_body_mutation
header_immediate_response - Redirect Example
basic_callout_server version - Basic server capabilities
jwt_auth - Using jwt authentication
client


callouts/go/extproc/internal/server/callout_server.go
```diff
@@ -0,0 +1,229 @@
...
+		if response != nil {
+			if err := stream.Send(response); err != nil {
+				return err
+			}
+		}
+	}
+}
+
+func (s *GRPCCalloutService) HandleRequestHeaders(headers *extproc.HttpHeaders) (*extproc.ProcessingResponse, error) {
+	return &extproc.ProcessingResponse{
```
Lessyless, line 184: This file doesn't have a line break after the function declarations, but the previous files did. Could you make it consistent?

pweiber:
@Lessyless Thanks for the initial review, I did the changes, let me know if I missed something or need new adjustments

callouts/go/extproc/examples/jwt_auth/jwt_auth_test.go
```diff
@@ -0,0 +1,171 @@
...
+
+	// Check if the error is a permission denied error
+	if got, want := err.Error(), "PermissionDenied"; !strings.Contains(got, want) {
+		t.Errorf("HandleRequestHeaders() error = %v, want %v", got, want)
+	}
+
+	// Check if the error message contains "Authorization token is invalid"
+	if got, want := err.Error(), "Authorization token is invalid"; !strings.Contains(got, want) {
+		t.Errorf("HandleRequestHeaders() error = %v, want %v", got, want)
+	}
```
Lessyless, line 170: I would suggest something similar to the following:
```
wantErr := status.Errorf(codes.PermissionDenied, "Authorization token is invalid")
if !cmp.Equal(err, wantErr) {
  // ...
}
```

Lessyless (commented):
Looks nice!

pweiber:
@jstraceski Should be good now, let me know if you find anything else

@Lessyless I also changed the tests from add_body and add_header to use the cmp, let me know if the cmp and the table-driven are looking good now

callouts/go/extproc/examples/client/client.go
```diff
@@ -0,0 +1,197 @@
...
+		reqBytes, err := json.Marshal(rawReq)
+		if err != nil {
+			return nil, err
+		}
+		req := &extproc.ProcessingRequest{}
+		if err := protojson.Unmarshal(reqBytes, req); err != nil {
+			return nil, err
+		}
+		requests = append(requests, req)
+	}
```
Lessyless, line 130: I wonder if we can replace this piece of code with the following?
```
	var rawRequests []json.RawMessage
	if err := json.Unmarshal([]byte(jsonData), &rawRequests); err != nil {
		return nil, err
	}

	var requests []*extproc.ProcessingRequest
	for _, rawReq := range rawRequests {
		req := &extproc.ProcessingRequest{}
		if err := protojson.Unmarshal(rawReq, req); err != nil {
			return nil, err
		}
		requests = append(requests, req)
	}
```

callouts/go/extproc/examples/client/client.go
```diff
@@ -0,0 +1,197 @@
...
+	if err != nil {
+		log.Fatalf("Failed to make JSON request: %v", err)
+	}
+
+	for _, resp := range responses {
+		respJSON, err := protojson.Marshal(resp)
+		if err != nil {
+			log.Fatalf("Failed to marshal response to JSON: %v", err)
+		}
+		log.Printf("Response: %s", string(respJSON))
```
Lessyless, line 195: Did I understand correctly that the purpose of this part is to print responses pretty? If yes, could we use [`prototext.Format`](https://pkg.go.dev/google.golang.org/protobuf/encoding/prototext)?

callouts/go/extproc/examples/client/client.go
```diff
@@ -0,0 +1,197 @@
...
+		log.Fatalf("Failed to serve gRPC: %v", err)
+	}
+}
+
+var (
+	tls        = flag.Bool("tls", false, "Connection uses TLS if true, else plain TCP")
+	certFile   = flag.String("cert_file", "", "The file containing the CA root cert file")
+	serverAddr = flag.String("addr", "localhost:8181", "The server address in the format of host:port")
+	dataJSON   = flag.String("data", "", "The JSON string containing the ProcessingRequest data")
+)
```
Lessyless, line 84: Could you move this group under the imports section (https://google.github.io/styleguide/go/decisions.html#flags).

Lessyless (commented):
Thank you for the update! Sorry for the confusion, but I meant something a bit different about using `cmp.Diff`. Please see the comments below.

I also took a look at the newly added `client.go` and `client_test.go` files. Just a couple of comments there. Overall good!

callouts/go/extproc/examples/jwt_auth/jwt_auth_test.go
Lessyless: This comment seems to still be relevant.

pweiber:
> Thank you for the update! Sorry for the confusion, but I meant something a bit different about using `cmp.Diff`. Please see the comments below.
> 
> I also took a look at the newly added `client.go` and `client_test.go` files. Just a couple of comments there. Overall good!

@Lessyless I believe it should be good now, let me know if this is what you expected for the `cmp.Diff` and `protocmp`.

For the jwt I had to make some changes based on the errors I was getting, changed the 'iat' and 'exp' to avoid scientific values, I was getting something like:

> s{header:{key:"decoded-exp"  raw_value:"1.720022207e+09"}},
> s{header:{key:"decoded-iat"  raw_value:"1.720018607e+09"}},

Also added the sortHeaders because the headers were showing in a different order than expected

Lessyless (commented):
Good! Please address the last part of comments, and that's all from my side.

callouts/go/extproc/examples/jwt_auth/jwt_auth_test.go
```diff
@@ -38,6 +25,13 @@ func generateTestJWTToken(privateKey []byte, claims jwt.MapClaims) (string, erro
 	return token.SignedString(privateKeyParsed)
 }
 
+// sortHeaders sorts headers in-place by their keys.
+func sortHeaders(headers []*base.HeaderValueOption) {
+	sort.Slice(headers, func(i, j int) bool {
+		return headers[i].Header.Key < headers[j].Header.Key
+	})
+}
+
```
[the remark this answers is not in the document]
pweiber: I tried some alternatives to use the `cmpopts.SortSlices` but I'm not sure if that would be the right approach, could you check If that makes sense? If I put the wantResponse and response directly it does not work, had to create the "Sorted"

```go
opts := []cmp.Option{
	protocmp.Transform(),
	cmpopts.SortSlices(func(x, y *base.HeaderValueOption) bool {
		return x.Header.Key < y.Header.Key
	}),
}

wantResponseSorted := wantResponse.Response.(*extproc.ProcessingResponse_RequestHeaders).RequestHeaders.Response.HeaderMutation.SetHeaders
responseSorted := response.Response.(*extproc.ProcessingResponse_RequestHeaders).RequestHeaders.Response.HeaderMutation.SetHeaders

if diff := cmp.Diff(wantResponseSorted, responseSorted, opts...); diff != "" {
	t.Errorf("HandleRequestHeaders() mismatch (-want +got):\n%s", diff)
}
```

Lessyless (approved):
Approved!

rshriram (approved):

PR #90, opened 2024-05-30
feat: Adding ext proc client example

jstraceski:

### DIFF
--- a/callouts/python/extproc/example/ext_proc_client.py
+++ b/callouts/python/extproc/example/ext_proc_client.py
@@ -28,18 +28,31 @@ from envoy.service.ext_proc.v3.external_processor_pb2_grpc import (
 )
 
 
+def _make_channel(
+  address: tuple[str, int], key: str | None = None
+) -> grpc.Channel:
+  addr_str = _addr_to_str(address)
+  if key:
+    with open(key, 'rb') as file:
+      creds = grpc.ssl_channel_credentials(file.read())
+      return grpc.secure_channel(addr_str, creds)
+  else:
+    return grpc.insecure_channel(addr_str)
+
+
 def make_json_request(
-  json_list: list[str], address: tuple[str, int]
+  json_list: list[str], address: tuple[str, int], key: str | None = None
 ) -> Iterator[ProcessingResponse]:
   """Make requests to a callout service with ProcessingRequest json data.
   Args:
     json_list: A list of json strings representing ProcessingRequest data.
     address: ip address of the callout server.
+    key: Local filepath to a chain authenication certificate.
   Returns:
     An iterator containg each response.
   """
   callouts = [Parse(data, ProcessingRequest()) for data in json_list]
-  with grpc.insecure_channel(_addr_to_str(address)) as channel:
+  with _make_channel(address, key) as channel:
     stub = ExternalProcessorStub(channel)
     for response in stub.Process(iter(callouts)):
       yield response
@@ -58,6 +71,15 @@ if __name__ == '__main__':
     help=('Address of the callout server in the format ip:port.'),
   )
   parser.add_argument(
+    '--cert',
+    type=str,
+    help=(
+      'Local path to a chain or root PEM authenication certificate.'
+      ' If specified, attemps to establish a secure connection to the callout'
+      ' server using the certificate.'
+    ),
+  )
+  parser.add_argument(
     '-d',
     '--data',
     type=str,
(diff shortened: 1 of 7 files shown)


callouts/python/extproc/example/ext_proc_client.py
```diff
@@ -0,0 +1,78 @@
...
+  ProcessingResponse,
+)
+from extproc.tests.basic_grpc_test import _addr_to_str
+from extproc.service.callout_tools import _addr
+from envoy.service.ext_proc.v3.external_processor_pb2_grpc import (
+  ExternalProcessorStub,
+)
+
+
+def make_json_request(
```
rshriram, line 31: make request. why is it json?

callouts/python/extproc/example/ext_proc_client.py
```diff
@@ -0,0 +1,78 @@
...
+) -> Iterator[ProcessingResponse]:
+  """Make requests to a callout service with ProcessingRequest json data.
+  Args:
+    json_list: A list of json strings representing ProcessingRequest data.
+    address: ip address of the callout server.
+  Returns:
+    An iterator containg each response.
+  """
+  callouts = [Parse(data, ProcessingRequest()) for data in json_list]
+  with grpc.insecure_channel(_addr_to_str(address)) as channel:
```
rshriram, line 42: need options for doing secure vs insecure

callouts/python/extproc/example/ext_proc_client.py
jstraceski: That's the format of the string that `Parse` takes in.

callouts/python/extproc/example/ext_proc_client.py
```diff
@@ -28,18 +28,31 @@
...
+
 def make_json_request(
-  json_list: list[str], address: tuple[str, int]
+  json_list: list[str], address: tuple[str, int], key: str | None = None
 ) -> Iterator[ProcessingResponse]:
   """Make requests to a callout service with ProcessingRequest json data.
   Args:
     json_list: A list of json strings representing ProcessingRequest data.
     address: ip address of the callout server.
+    key: Local filepath to a chain authenication certificate.
```
rshriram, line 50: Lets not use the Cert chain as a way of saying tls or not. we configure the client to accept unverified certs. Cert chain is overkill. We have a boolean (-plaintext) that allows us to pick plaintext or tls.

callouts/python/extproc/example/ext_proc_client.py
```diff
@@ -28,18 +28,31 @@
...
   Returns:
     An iterator containg each response.
   """
   callouts = [Parse(data, ProcessingRequest()) for data in json_list]
-  with grpc.insecure_channel(_addr_to_str(address)) as channel:
+  with _make_channel(address, key) as channel:
     stub = ExternalProcessorStub(channel)
     for response in stub.Process(iter(callouts)):
```
rshriram, line 57: So the user could have a series of request_headers as messages? Will they be able to exercise how their callout server handles stream closure given that envoy creates a new stream for each individual client request?

callouts/python/extproc/example/ext_proc_client.py
```diff
@@ -72,7 +94,12 @@ def make_json_request(
     'prints out the responses.'
   )
   args = parser.parse_args()
-  # Preform callouts and print the responses.
-  logging.info(
-    list(make_json_request(json_list=args.data, address=args.address))
+  # Preform callouts and collect the responses.
+  responses = list(
+    make_json_request(json_list=args.data, address=args.address, key=args.key)
```
rshriram, line 99: can we go one step further as follows:
1. Have a pre-created dummy request with headers. This way, the user can simply run the client with just the server IP as the arg
2. Allow the user to specify the ProcessingRequest proto/json as an args

rshriram (approved):

callouts/python/extproc/example/ext_proc_client.py
jstraceski: I can make a follow-up pr to address this, technically they can provide either the cert chain or root cert.

callouts/python/extproc/example/ext_proc_client.py
jstraceski: The requests for this client should be able to handle individual stream closures, if the server closes one request it will close all requests though. The user would need to make multiple calls to the docker binary if they want to test multiple closures.

callouts/python/extproc/example/ext_proc_client.py
jstraceski: 1) Noted, I can add a dummy request in a followup 
2) Is that not what `-d` would be doing?

PR #91, opened 2024-06-06
feat: Read serialized HTTP1 headers from string or file

martijneken:
Uses the Quiche Balsa library to parse HTTP1 headers and converts 
them into HTTP2 headers like ext_proc servers would see.

Also includes a large number of minor features and fixes:

- A testing showcase sample that exercises most test runner features
- An option to set the (fixed) host timestamp + tests for static time
- Generic string matching on serialized headers (key: value)
- Improve string matcher output format for clarity
- Supporting reading test files from relative paths
- A `--nobench` flag to turn off all benchmarking (faster iteration)
- Small fix to `--loglevel` flag overrides

Checks:

- [x] Tests pass

### DIFF
--- a/plugins/README.md
+++ b/plugins/README.md
@@ -15,8 +15,8 @@ We recommend the following process:
     [Proxy-Wasm](https://github.com/proxy-wasm) SDKs as a starting point, write
     a wasm plugin in a language of your choice. Get it building.
 1.  Write a plugin test file (textproto) to specify the plugin's functional
-    expectations ([example](samples/config_denylist/tests.textpb)). Consult the
-    plugin tester [proto API](test/runner.proto) as needed.
+    expectations ([example](samples/testing/tests.textpb)). Consult the plugin
+    tester [proto API](test/runner.proto) as needed.
 1.  Add `benchmark: true` to tests that exemplify common wasm operations
     ([example](samples/add_header/tests.textpb)).
 1.  Run + Test + Benchmark your wasm plugin as follows!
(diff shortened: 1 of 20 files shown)


plugins/README.md
```diff
@@ -31,15 +31,18 @@ docker run -it -v $(pwd):/mnt \
 Tips:
 
 -   When benchmarking and publishing, compile a release (optimized) wasm build.
--   To see a trace of logs and wasm ABI calls, add `--loglevel=TRACE`.
 -   Try sending empty or invalid input. Verify your plugin doesn't crash.
+-   To see plugin-emitted logs on the console, add `--logfile=/dev/stdout`.
```
dallen68, line 35: I think we should specify that the complete flag is --test_arg=--loglevel=TRACE.

dallen68 (approved):

plugins/test/dynamic_test.cc
```diff
@@ -370,20 +394,104 @@ void DynamicTest::FindString(const std::string& phase, const std::string& type,
...
+        headers.RemoveAllOfHeader("Host");
+      }
+    }
+  } else {
+    hdrs.InsertOrAppend(":status", headers.response_code());
+  }
+
+  // Emit normal headers to map, coalescing as we go.
+  for (const auto& [key, value] : headers.lines()) {
+    hdrs.InsertOrAppend(key, value);
```
martijneken, line 458: TODO: lowercase the key here, like Envoy does

martijneken: Done

plugins/README.md
martijneken: That's how you set flags under `bazel test`. However most users will be interacting with the tool directly, so `--loglevel` is the flag they'd use.

PR #105, opened 2024-07-22
feat: create a C++ sample for overwrite header plugin.

walves-cit:
This plugin is a overwriting request and response header showcase.

Technically, this is performed by replacing some existing header according to the desired key.
 

plugins/samples/overwrite_header/tests.textpb
```diff
@@ -0,0 +1,37 @@
+# Expect no matches, so no overwrite.
+test {
+  name: "NoOverwrite"
+  benchmark: true
```
martijneken, line 4: nit: Probably not worth benchmarking these tests.

plugins/samples/overwrite_header/plugin.cc
```diff
@@ -0,0 +1,39 @@
...
+// [START serviceextensions_plugin_overwrite_header]
+#include "proxy_wasm_intrinsics.h"
+
+class MyHttpContext : public Context {
+ public:
+  explicit MyHttpContext(uint32_t id, RootContext* root) : Context(id, root) {}
+
+  FilterHeadersStatus onRequestHeaders(uint32_t headers,
+                                       bool end_of_stream) override {
+    // Change the key and value according to your needs
```
martijneken, line 24: Please point out in the comment how this differs from addRequestHeader (I had to go look it up). The difference is that addRequestHeader will append a value if one already exists (e.g. key: value -> key: value, value2) whereas this operation replaces the value entirely (key: value2).

You might want to show that difference in behavior by calling addResponseHeader on the response path below.

plugins/samples/overwrite_header/tests.textpb
```diff
@@ -0,0 +1,37 @@
+# Expect no matches, so no overwrite.
+test {
+  name: "NoOverwrite"
```
martijneken, line 3: Note that the new headers (RequestHeader/ResponseHeader) still get added in this test. Please add the corresponding expectations to result{}.

plugins/samples/overwrite_header/tests.textpb
```diff
@@ -0,0 +1,37 @@
...
+# Expect the plugin to overwrite the value for the response header ResponseHeader.
+test {
+  name: "OverwriteResponseHeader"
+  benchmark: true
+  response_headers {
+    input { 
+      header { key: "ResponseHeader" value: "SomeHeaderValue" }
+      header { key: "OtherHeader" value: "OtherValue" }
+    }
+    result { has_header { key: "ResponseHeader" value: "changed" } }
```
martijneken, line 35: Per other suggestion, if you instead call addResponseHeader on the response path, you can test it here:

`result { has_header { key: "ResponseHeader" value: "SomeHeaderValue, changed" } }`

plugins/samples/overwrite_header/tests.textpb
```diff
@@ -0,0 +1,37 @@
...
+# Expect the plugin to overwrite the value for the request header RequestHeader.
+test {
+  name: "OverwriteRequestHeader"
+  benchmark: true
+  request_headers {
+    input { 
+      header { key: "RequestHeader" value: "SomeHeaderValue"}
+      header { key: "OtherHeader" value: "OtherValue" } 
+    }
+    result { has_header { key: "RequestHeader" value: "changed" } }
```
martijneken, line 23: For clarity you might show that the OtherHeader still exists. Same in the test below.

plugins/samples/overwrite_header/tests.textpb
```diff
@@ -0,0 +1,37 @@
...
+  }
+}
+# Expect the plugin to overwrite the value for the response header ResponseHeader.
+test {
```
martijneken, line 27: You could combine this test with "OverwriteRequestHeader" (so one stream that executes both callbacks), or split "NoOverwrite" test into two. Worth being consistent.

plugins/samples/overwrite_header/BUILD
```diff
@@ -0,0 +1,19 @@
+load("//:plugins.bzl", "proxy_wasm_plugin_cpp", "proxy_wasm_plugin_rust", "proxy_wasm_tests")
+
+licenses(["notice"])  # Apache 2
+
+proxy_wasm_plugin_cpp(
```
martijneken, line 5: If possible I'd like to get both C++ and Rust tests for each sample. The Rust plugin will be very similar to this one: https://github.com/GoogleCloudPlatform/service-extensions-samples/blob/main/plugins/samples/add_header/plugin.rs

plugins/samples/overwrite_header/tests.textpb
walves-cit: I did not understand. The headers mentioned by you aren't added here. So no changes are expected here.

plugins/samples/overwrite_header/tests.textpb
walves-cit: Spliting.

plugins/samples/overwrite_header/tests.textpb
martijneken: Ah, my point was that these lines are always called:
```
# request path
replaceRequestHeader("RequestHeader", "changed");
# response path
replaceResponseHeader("ResponseHeader", "changed");
```
Regardless of whether the input has "RequestHeader" or "ResponseHeader", this code will set those values into the output. So IMO the "NoOverwrite" test should demonstrate that the headers always get added, by verifying this in the result{} stanza:
```
    result { 
      has_header { key: "RequestHeader" value: "changed" }
    }
```

---
What you have written: `result { no_header {value: "changed"} }` doesn't test what you expect. Because the key is unspecified, it checks that there's no header called "" (always true I expect). I can fix/clarify this in the test framework.

martijneken: https://github.com/GoogleCloudPlatform/service-extensions-samples/pull/106 to fix

walves-cit: Now I got it... This plugin is intended to only overwrite and not to add if its missing.
I've just made the corresponding changes to the code and tests.
Pls take a look.

plugins/samples/overwrite_header/BUILD
martijneken: Withdrawn.

plugins/samples/overwrite_header/tests.textpb
```diff
@@ -1,37 +1,47 @@
 # Expect no matches, so no overwrite.
 test {
-  name: "NoOverwrite"
-  benchmark: true
+  name: "NoOverwriteRequestHeader"
+  benchmark: false
   request_headers {
     input { header { key: "SomeRequestHeaderKey" value: "SomeHeaderValue" } }
-    result { has_header { key: "SomeRequestHeaderKey" value: "SomeHeaderValue" } }
+    result { no_header {key: "RequestHeader" value: "changed"} }
```
martijneken, line 7: `value` is ignored here, can omit (same in next test)

plugins/samples/overwrite_header/plugin.cc
```diff
@@ -15,21 +15,32 @@
...
+// this plugin overwrites the entire value for the specified key.
 class MyHttpContext : public Context {
  public:
   explicit MyHttpContext(uint32_t id, RootContext* root) : Context(id, root) {}
 
   FilterHeadersStatus onRequestHeaders(uint32_t headers,
                                        bool end_of_stream) override {
     // Change the key and value according to your needs
-    replaceRequestHeader("RequestHeader", "changed");
+    auto header_key = "RequestHeader";
```
martijneken, line 28: const auto    (everywhere possible)

plugins/samples/overwrite_header/plugin.cc
```diff
@@ -15,21 +15,32 @@
...
 
   FilterHeadersStatus onResponseHeaders(uint32_t headers,
                                         bool end_of_stream) override {
     // Change the key and value according to your needs
-    replaceResponseHeader("ResponseHeader", "changed");
+    auto header_key = "ResponseHeader";
+    auto header = getResponseHeader(header_key);
+    if (header->size() > 0) {
+      replaceResponseHeader(header_key, "changed");
+    }
```
martijneken, line 43: Works fine, but this identical example doesn't really teach the reader anything new. This might be an opportunity to call replaceResponseHeader without the existence check, and show (in a comment and in the tests) how that behavior is different (adds the header if it's missing).

The fact that "replace" refers to the value and not the key is the most surprising behavior of replace*Header(), so I think we should call it out clearly in the sample.

plugins/samples/overwrite_header/tests.textpb
martijneken: Thanks for fixing

plugins/samples/overwrite_header/tests.textpb
```diff
@@ -1,37 +1,47 @@
 # Expect no matches, so no overwrite.
 test {
-  name: "NoOverwrite"
-  benchmark: true
+  name: "NoOverwriteRequestHeader"
```
martijneken, line 3: naming nit: You might call this DontAddRequestHeader for clarity

If you change the response path behavior as I suggested, you could call that one DoAddResponseHeader

plugins/samples/overwrite_header/tests.textpb
```diff
@@ -1,37 +1,47 @@
...
+  benchmark: false
   request_headers {
     input { header { key: "SomeRequestHeaderKey" value: "SomeHeaderValue" } }
```
martijneken, line 6: nit: If you want to add other headers into the mix (SomeRequestHeaderKey), you should show that they survive unchanged by testing in `result{}` that they still have the same value.

plugins/samples/overwrite_header/plugin.cc
martijneken: You've updated both, but I was suggesting that one handler (e.g. request) does the existence check, whereas the other (e.g. response) does a blind replace. That way users can see the difference.

walves-cit: I see... changing now

walves-cit: Done

martijneken: Thanks!

martijneken:
LGTM, waiting for CI to confirm

martijneken:
No idea why header-check stopped working. Bypassing to merge.

martijneken (approved):

PR #106, opened 2024-07-31
fix: Plugin testing sharp edges

martijneken:
Various small fixes

- error when header key is omitted
- don't benchmark stream teardown

### DIFF
--- a/plugins/test/dynamic_test.cc
+++ b/plugins/test/dynamic_test.cc
@@ -294,33 +294,37 @@ void DynamicTest::BenchHttpHandlers(benchmark::State& state) {
   for (const auto& response_body : cfg_.response_body()) {
     response_body_chunks.emplace_back(response_body.input().content());
   }
+  std::optional<TestHttpContext> stream;
   for (auto _ : state) {
+    // Pausing timing here is not recommended. One way we could avoid it:
+    // - include stream context create/destroy cost in handler benchmarks
+    // - don't hand ownership of body chunks to stream context
     state.PauseTiming();
-    auto stream = TestHttpContext(handle);
+    stream.emplace(handle);  // create/destroy TestHttpContext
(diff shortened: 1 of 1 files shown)


plugins/test/dynamic_test.cc
```diff
@@ -320,6 +320,7 @@ void DynamicTest::BenchHttpHandlers(benchmark::State& state) {
       benchmark::DoNotOptimize(res);
       BM_RETURN_IF_FAILED(handle);
     }
+    state.PauseTiming();
```
leonm1, line 323: The benchmarking framework's documentation says this should be avoided in loops: https://github.com/google/benchmark/blob/25e5c52a112a56acdc37cace73025f7327f03be7/include/benchmark/benchmark.h#L805

Does this make a measurable difference in the outcome?

leonm1 (approved):

plugins/test/dynamic_test.cc
martijneken: Ah, interesting guidance, I hadn't seen that. The problem is that right now we call PauseTiming/ResumeTiming above, which excludes TestHttpContext construction cost. But then the TestHttpContext teardown cost is included here, so we are including half of BenchStreamLifecycle in this benchmark.

I would be OK with one model or the other, but not halfway, hence this quick fix.

Long term I think I'd prefer to get rid of these calls (thus including the TestHttpContext lifecycle costs). To do that, we'd also have to stop handing ownership of BODY chunks to SendRequestBody. @mpwarres FYI

martijneken: Heh, but at the same time this seems to cause asan/tsan failures, presumably on the last iteration... `StopTimer: Check 'running_' failed.`

martijneken: Fixed a different way (std::optional stack variable outside the benchmark loop). Left a comment about possible future work.

PR #107, opened 2024-08-08
feat: create a C++ sample for normalize HTTP header plugin.

walves-cit:
This plugin is a normalize request header showcase.

Technically, this is performed by creating a new HTTP header `client-device-type` to shard requests based on device according to the existence of the `HTTP Client Hints` header or the `User-Agent` value set.


plugins/README.md
```diff
@@ -60,6 +60,9 @@ for your own plugin. Extend them to fit your particular use case.
     expressions, compiling them at plugin initialization.
 *   [Overwrite HTTP request & response headers](samples/overwrite_header):
     Overwrite a header on both the client request and server response paths.
+*   [Normalize a HTTP header on request](samples/normalize_header): Check the host
```
mpwarres, line 63: Device classification isn't typically done using the "host" header (which just reflects the authority portion of the URL). Instead in the past it was done based on the "User-agent" request header, which is starting to be replaced by HTTP Client Hints: [https://developer.chrome.com/docs/privacy-security/user-agent-client-hints](https://www.google.com/url?q=https://developer.chrome.com/docs/privacy-security/user-agent-client-hints&sa=D&source=docs&ust=1723153515044619&usg=AOvVaw2UuOZfnAjN_eUX6sfCP9I9)

I'd propose that instead we have the plugin implement onRequestHeaders to do the following:

* if the "Sec-CH-UA-Mobile" request header is present and has value "?1", then set "client-device-type" header to "mobile".
* Otherwise, if the "User-Agent" request header is present and contains the substring "mobile" (case insensitive), then set "client-device-type" header to "mobile"
* Otherwise, set "client-device-type" request header to "unknown".

mpwarres (changes requested):

plugins/README.md
walves-cit: Done.

plugins/README.md
```diff
@@ -60,6 +60,9 @@ for your own plugin. Extend them to fit your particular use case.
     expressions, compiling them at plugin initialization.
 *   [Overwrite HTTP request & response headers](samples/overwrite_header):
     Overwrite a header on both the client request and server response paths.
+*   [Normalize a HTTP header on request](samples/normalize_header): Creates a new
+    HTTP header (client-device-type) to shard requests based on device according
+    to the existence of the HTTP Client Hints header or the User-Agent value set.
```
mpwarres, line 65: suggest minor wording tweak: "according to the existence of HTTP Client Hints or User-Agent header values."

plugins/samples/normalize_header/plugin.cc
```diff
@@ -0,0 +1,51 @@
...
+// Determines client device type based on request headers.
+class MyHttpContext : public Context {
+ public:
+  explicit MyHttpContext(uint32_t id, RootContext* root) : Context(id, root) {}
+
+  FilterHeadersStatus onRequestHeaders(uint32_t headers,
+                                       bool end_of_stream) override {
+    // Check "Sec-CH-UA-Mobile" header first (highest priority)
+    const auto mobile_header = getRequestHeader("Sec-CH-UA-Mobile");
+    if (mobile_header && mobile_header->toString() == "?1") {
```
mpwarres, line 28: Using `mobile_header->view()` instead of `mobile_header->toString()` avoids a string copy, so is preferable

plugins/samples/normalize_header/plugin.cc
```diff
@@ -0,0 +1,51 @@
...
+      addRequestHeader("client-device-type", "mobile");
+      return FilterHeadersStatus::Continue;
+    }
+
+    // Check "User-Agent" header for mobile substring (case insensitive)
+    const auto user_agent = getRequestHeader("User-Agent");
+    if (user_agent) {
+      const auto user_agent_str = user_agent->toString();
```
mpwarres, line 36: Can this use absl::StrContainsIgnoreCase? i.e.

```
if (absl::StrContainsIgnoreCase(user_agent->view(), "mobile")) {
  // ...
}
```

mpwarres (commented):
LGTM once remaining comments (all minor) addressed. Thanks!

plugins/samples/normalize_header/plugin.cc
mpwarres: Looks good! Can you remove the `const auto user_agent_str = user_agent->toString();` inside the if block, which is no longer used?

walves-cit: My bad... removed.

mpwarres (approved):
LGTM, thanks!

walves-cit:
Hi @mpwarres, is it normal for the merge to get stuck waiting the header-check?

mpwarres:
> Hi @mpwarres, is it normal for the merge to get stuck waiting the header-check?

I don't think so. Your branch needs to be updated to pull in latest changes from `main` anyways, how about doing that and then we can re-kick the CI?

walves-cit:
> > Hi @mpwarres, is it normal for the merge to get stuck waiting the header-check?
> 
> I don't think so. Your branch needs to be updated to pull in latest changes from `main` anyways, how about doing that and then we can re-kick the CI?

Done

mpwarres:
@walves-cit this can be merged now

PR #111, opened 2024-08-13
feat: create a C++ sample for block request with particular URL/header plugin

walves-cit:
This plugin is a block request with particular URL/header showcase.

Technically, this is performed checking if the client's referer matches an expected domain. If there's a mismatch, a 403 Forbidden error is returned.


plugins/samples/block_request/plugin.cc
```diff
@@ -0,0 +1,57 @@
...
+class MyHttpContext : public Context {
+ public:
+  explicit MyHttpContext(uint32_t id, RootContext* root) : Context(id, root) {}
+
+  FilterHeadersStatus onRequestHeaders(uint32_t headers,
+                                       bool end_of_stream) override {
+    const auto referer = getRequestHeader("Referer");
+    const auto host = getRequestHeader("Host");
+    // Check if referer and host match.
+    if (referer && host && !absl::StrContains(referer->view(), host->view())) {
```
mpwarres, line 32: Per earlier comment re: matching against a single, allowed domain, I'd suggest defining a constant for an allowed Referer value early in this file:

```
constexpr absl::string_view kAllowedReferer = "example.com";
```
and then modifying the logic here to be:
```
if (!referer || !absl::StrContains(referer->view(), kAllowedReferer)) {
  // send 403 logic
}
```

plugins/samples/block_request/plugin.cc
```diff
@@ -0,0 +1,57 @@
...
+      const auto requestId = generateRandomRequestId();
+      sendLocalResponse(403, "", "Forbidden - Request ID: " + requestId, {});
+      LOG_INFO("Forbidden - Request ID: " + requestId);
+      return FilterHeadersStatus::StopAllIterationAndWatermark;
+    }
+
+    addRequestHeader("Allowed", "true");
```
mpwarres, line 39: Suggest choosing a header name such as "my-plugin-allowed" to make clear that this is not a standard HTTP header, and is added here just for illustrative purposes.

plugins/README.md
```diff
@@ -63,6 +63,9 @@ for your own plugin. Extend them to fit your particular use case.
 *   [Normalize a HTTP header on request](samples/normalize_header): Creates a new
     HTTP header (client-device-type) to shard requests based on device according
     to the existence of HTTP Client Hints or User-Agent header values.
+*   [Block request with particular URL/header](samples/block_request): Check if the
```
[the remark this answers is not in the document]
walves-cit: Makes sense

plugins/samples/block_request/plugin.cc
walves-cit: Done.

plugins/samples/block_request/plugin.cc
walves-cit: Changed

mpwarres (approved):

PR #112, opened 2024-08-14
feat: create a C++ sample plugin for HTTP redirect

walves-cit:
This plugin is a HTTP redirect showcase.

Technically, this is performed by making an HTTP redirect to another URL depending on the original URL requested.

### DIFF
--- a/plugins/README.md
+++ b/plugins/README.md
@@ -60,6 +60,9 @@ for your own plugin. Extend them to fit your particular use case.
     expressions, compiling them at plugin initialization.
 *   [Overwrite HTTP request & response headers](samples/overwrite_header):
     Overwrite a header on both the client request and server response paths.
+*   [Normalize a HTTP header on request](samples/normalize_header): Creates a new
+    HTTP header (client-device-type) to shard requests based on device according
+    to the existence of HTTP Client Hints or User-Agent header values.
 
 # Samples tests
 
--- a/plugins/README.md
+++ b/plugins/README.md
@@ -63,6 +63,8 @@ for your own plugin. Extend them to fit your particular use case.
 *   [Normalize a HTTP header on request](samples/normalize_header): Creates a new
     HTTP header (client-device-type) to shard requests based on device according
     to the existence of HTTP Client Hints or User-Agent header values.
+*   [Perform a HTTP redirect](samples/redirect): For a given URL makes a redirect
+    to another URL.
 
 # Samples tests
 
--- a/plugins/README.md
+++ b/plugins/README.md
@@ -63,8 +63,7 @@ for your own plugin. Extend them to fit your particular use case.
 *   [Normalize a HTTP header on request](samples/normalize_header): Creates a new
     HTTP header (client-device-type) to shard requests based on device according
     to the existence of HTTP Client Hints or User-Agent header values.
-*   [Perform a HTTP redirect](samples/redirect): For a given URL makes a redirect
-    to another URL.
+*   [Perform a HTTP redirect](samples/redirect): Redirect a given URL to another URL.
 
 # Samples tests
 
--- a/plugins/README.md
+++ b/plugins/README.md
@@ -63,6 +63,8 @@ for your own plugin. Extend them to fit your particular use case.
 *   [Normalize a HTTP header on request](samples/normalize_header): Creates a new
     HTTP header (client-device-type) to shard requests based on device according
     to the existence of HTTP Client Hints or User-Agent header values.
+*   [Overwrite origin response error code](samples/overwrite_errcode): Overwrites
+    error code served from origin from 5xx error to 4xx error class.
 *   [Perform a HTTP redirect](samples/redirect): Redirect a given URL to another URL.
 
 # Samples tests
--- a/plugins/samples/redirect/plugin.cc
+++ b/plugins/samples/redirect/plugin.cc
@@ -13,7 +13,7 @@
 // limitations under the License.
 
 // [START serviceextensions_plugin_redirect]
-#include "absl/strings/match.h"
+#include "absl/strings/str_replace.h"
 #include "proxy_wasm_intrinsics.h"
 
 // This sample may redirects the request for other URL according to the URL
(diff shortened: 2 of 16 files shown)


plugins/README.md
```diff
@@ -63,6 +63,8 @@ for your own plugin. Extend them to fit your particular use case.
 *   [Normalize a HTTP header on request](samples/normalize_header): Creates a new
     HTTP header (client-device-type) to shard requests based on device according
     to the existence of HTTP Client Hints or User-Agent header values.
+*   [Perform a HTTP redirect](samples/redirect): For a given URL makes a redirect
```
mpwarres, line 66: "Redirect a given URL to another URL"

plugins/samples/redirect/plugin.cc
```diff
@@ -0,0 +1,43 @@
...
+
+// This sample may redirects the request for other URL according to the URL
+// requested.
+class MyHttpContext : public Context {
+ public:
+  explicit MyHttpContext(uint32_t id, RootContext* root) : Context(id, root) {}
+
+  FilterHeadersStatus onRequestHeaders(uint32_t headers,
+                                       bool end_of_stream) override {
+    const auto location = getRequestHeader("Location");
```
mpwarres, line 27: This doesn't look quite right to me: the "Location" response header is used to indicate the target of a redirect (which the code does set, later on), but here, the code is looking for a "Location" *request header*, which won't exist because clients don't send "Location".

Backing up, I think a good example for a redirect would be mapping all requests whose path starts with "/foo/..." to instead go to "/bar/...", so https://example.com/foo/pictures/cat.jpg would get redirected to https://example.com/bar/pictures/cat.jpg.

So to do this you could have logic that matches the beginning of the ":path" header against some fixed string (in the above case, "/foo"), and if there is a match, replaces the matched section with another string ("/bar" above) and uses that string to construct the "Location" response header to send in the 301 response.

walves-cit: Make sense. Done

plugins/README.md
walves-cit: Done

plugins/samples/redirect/plugin.cc
```diff
@@ -0,0 +1,40 @@
...
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+// [START serviceextensions_plugin_redirect]
+#include "absl/strings/str_replace.h"
+#include "proxy_wasm_intrinsics.h"
+
+// This sample may redirects the request for other URL according to the URL
```
mpwarres, line 19: "This sample redirects any requests to paths starting with /foo/ to instead use path /bar/."

plugins/samples/redirect/plugin.cc
```diff
@@ -0,0 +1,40 @@
...
+// requested.
+class MyHttpContext : public Context {
+ public:
+  explicit MyHttpContext(uint32_t id, RootContext* root) : Context(id, root) {}
+
+  FilterHeadersStatus onRequestHeaders(uint32_t headers,
+                                       bool end_of_stream) override {
+    auto path = getRequestHeader(":path")->toString();
+    // Change the values according to your needs.
+    if (absl::StrReplaceAll({{"/foo/images", "/bar/images"}}, &path) > 0) {
```
mpwarres, line 29: I don't think this is fully correct, since it would map "/qux/foo/images" to /qux/bar/images", when we only want to rewrite paths under top-level directory "/foo/..." to be "/bar/...".

Suggest instead doing:

```
constexpr std::string_view old_path_prefix = "/foo/";
constexpr std::string_view new_path_prefix = "/bar/";
if (absl::StartsWith(path, old_path_prefix)) {
  std::string new_path = absl::StrCat(new_path_prefix, path.substr(old_path_prefix.length()));
  sendLocalResponse(
    301, "", absl::StrCat("Content moved to ", new_path),
    {{"Location", new_path}});
    return FilterHeadersStatus::ContinueAndEndStream;
}
```

plugins/samples/redirect/plugin.cc
```diff
@@ -0,0 +1,40 @@
...
+      sendLocalResponse(301, "", "", {{"Location", path}});
+
+      return FilterHeadersStatus::StopAllIterationAndWatermark;
```
mpwarres, line 32: When a plugin onRequestHeaders() callback generates a local response and is not going to process the stream further, it should return `FilterHeadersStatus::ContinueAndEndStream`.

I noticed that there may be a few other examples that are returning `FilterHeadersStatus::StopAllIterationAndWatermark` after generating a local response, these should also be changed to return `FilterHeadersStatus::ContinueAndEndStream` (though maybe in a separate PR).

I don't think there are any examples in the current set that need to return return `FilterHeadersStatus::StopAllIterationAndWatermark`. (Which is good, because we don't really support that yet, and basically ignore it.)

plugins/samples/redirect/tests.textpb
```diff
@@ -0,0 +1,25 @@
...
+    input { header { key: ":path" value: "/main/somepage/otherpage" } } 
+    result { 
+      has_header { key: ":path" value: "/main/somepage/otherpage" }
+      no_header { key: "Location" }
+    }
+  }
+}
+# With expected path, expected imediate with http status code 301 and
+# redirect to other location.
+test {
```
mpwarres, line 15: Can you add a test case that ensures that "/foo/" is only replaced with "/bar/" when at the beginning of the URL path?

plugins/samples/redirect/plugin.cc
walves-cit: Got it... Gonna change this in all other open PRs. Tks

plugins/samples/redirect/plugin.cc
walves-cit: Makes sense... Done

plugins/samples/redirect/tests.textpb
walves-cit: Done

mpwarres (approved):
LG, thanks!

PR #113, opened 2024-08-28
feat: create a C++ sample plugin for JWT authorization.

walves-cit:
This plugin is a JWT authorization showcase.

Technically, this is performed by ensuring that the request query string has a valid signed JWT. Case valid it will be removed from the URL.


plugins/samples/jwt_auth/plugin.cc
```diff
@@ -0,0 +1,95 @@
...
+  explicit MyRootContext(uint32_t id, std::string_view root_id)
+      : RootContext(id, root_id) {}
+
+  bool onConfigure(size_t config_len) override {
+    // Fetch plugin config and take ownership of the buffer.
+    config_ =
+        getBufferBytes(WasmBufferType::PluginConfiguration, 0, config_len);
+
+    // Read the RSA key from the config file.
+    const auto rsa_key = config_->toString();
```
mpwarres, line 33: Doesn't hurt to be explicit about type here, i.e. `const std::string rsa_key = ...;`

plugins/samples/jwt_auth/plugin.cc
```diff
@@ -0,0 +1,95 @@
...
+    jwks = google::jwt_verify::Jwks::createFrom(
+        rsa_key, google::jwt_verify::Jwks::Type::PEM);
+    return jwks->getStatus() == google::jwt_verify::Status::Ok;
+  }
+
+  google::jwt_verify::JwksPtr jwks;
```
mpwarres, line 39: Suggest making the field private, and either defining an accessor method, or sharing a pointer to it with MyHttpContext via an added MyHttpContext constructor param.

mpwarres (commented):
LGTM mod comments

plugins/samples/jwt_auth/plugin.cc
walves-cit: Done

plugins/samples/jwt_auth/plugin.cc
```diff
@@ -0,0 +1,96 @@
...
+    jwks = google::jwt_verify::Jwks::createFrom(
+        rsa_key, google::jwt_verify::Jwks::Type::PEM);
+    return jwks->getStatus() == google::jwt_verify::Status::Ok;
+  }
+
+  const google::jwt_verify::JwksPtr& getJwks() const { return jwks; }
+
+ private:
+  WasmDataPtr config_;
+  google::jwt_verify::JwksPtr jwks;
```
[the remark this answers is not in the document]
walves-cit: Done

mpwarres (approved):

PR #114, opened 2024-09-04
feat: create a C++ sample plugin for HMAC token authorization.

walves-cit:
This plugin is a HMAC token authorization showcase.

Technically, this is performed by ensuring that the request query string has a valid HMAC token.


plugins/samples/hmac_authtoken/plugin.cc
```diff
@@ -0,0 +1,85 @@
...
+// Replace with your desired secret key.
+const std::string kSecretKey = "your_secret_key";
+
+class MyHttpContext : public Context {
+ public:
+  explicit MyHttpContext(uint32_t id, RootContext* root) : Context(id, root) {}
+
+  FilterHeadersStatus onRequestHeaders(uint32_t headers,
+                                       bool end_of_stream) override {
+    boost::system::result<boost::urls::url> url =
```
mpwarres, line 35: Would be good to handle the case where the boost::system::result contains an error, i.e. has_value() returns false (or equivalently, has_error() returns true).

This applies to a few other existing plugins as well, e.g. plugins/samples/jwt_auth/plugin.cc and plugins/samples/ab_testing/plugin.cc, which both dereference a result<url> without first checking if it is non-error. I'm guessing that would trigger an exception or undefined behavior.

plugins/samples/hmac_authtoken/plugin.cc
```diff
@@ -0,0 +1,85 @@
...
+        boost::urls::parse_relative_ref(getRequestHeader(":path")->toString());
+    auto it = url->params().find("token");
+    // Check if the HMAC token exists.
+    if (it == url->params().end()) {
+      LOG_INFO("Access forbidden - missing token.");
+      sendLocalResponse(403, "", "Access forbidden - missing token.\n", {});
+      return FilterHeadersStatus::ContinueAndEndStream;
+    }
+
+    const auto token = (*it).value;
```
mpwarres, line 45: Expand `auto` into actual type, since it's not obvious from the expression alone

plugins/samples/hmac_authtoken/plugin.cc
```diff
@@ -0,0 +1,85 @@
...
+    // Strip the token from the URL.
+    url->params().erase(it);
+    const auto path = url->buffer();
```
mpwarres, line 48: Expand `auto` into actual type

plugins/samples/hmac_authtoken/plugin.cc
```diff
@@ -0,0 +1,85 @@
...
+      return FilterHeadersStatus::ContinueAndEndStream;
+    }
+
+    replaceRequestHeader(":path", path);
+    return FilterHeadersStatus::Continue;
+  }
+
+ private:
+  // Helper function to convert binary data to a hexadecimal string.
+  std::string toHexString(const unsigned char* data, size_t length) {
```
mpwarres, line 63: Could absl::BytesToHexString() be used instead?

If this method needs to be retained, suggest passing `data` and `length` as a `std::string_view`

plugins/samples/hmac_authtoken/plugin.cc
walves-cit: Got it... Added a URL parsing error check.

plugins/samples/hmac_authtoken/plugin.cc
walves-cit: Done

plugins/samples/hmac_authtoken/plugin.cc
walves-cit: Done

plugins/samples/hmac_authtoken/plugin.cc
```diff
@@ -0,0 +1,82 @@
...
+
+    auto it = url->params().find("token");
+    // Check if the HMAC token exists.
+    if (it == url->params().end()) {
+      LOG_INFO("Access forbidden - missing token.");
+      sendLocalResponse(403, "", "Access forbidden - missing token.\n", {});
+      return FilterHeadersStatus::ContinueAndEndStream;
+    }
+
+    const std::string token = (*it).value;
```
mpwarres, line 52: nit: can this use `it->value`?

mpwarres (approved):

plugins/samples/hmac_authtoken/plugin.cc
mpwarres: Oh, interesting--thanks for the explanation!

PR #117, opened 2024-09-09
feat: create a C++ sample plugin for HMAC cookie authorization.

walves-cit:
This plugin is a HMAC cookie authorization showcase.

Technically, this is performed by ensuring that the request has a valid HMAC cookie.

### DIFF
--- a/plugins/samples/hmac_authcookie/BUILD
+++ b/plugins/samples/hmac_authcookie/BUILD
@@ -1,0 +1,27 @@
+load("//:plugins.bzl", "proxy_wasm_plugin_cpp", "proxy_wasm_plugin_rust", "proxy_wasm_tests")
+
+licenses(["notice"])  # Apache 2
+
+proxy_wasm_plugin_cpp(
+    name = "plugin_cpp.wasm",
+    srcs = ["plugin.cc"],
+    deps = [
+        "@proxy_wasm_cpp_sdk//contrib:contrib_lib",
+        "@com_google_absl//absl/strings",
+        "@boringssl//:ssl",
+    ],
+    linkopts = [
+        # To avoid the error:
+        # library_pthread.js:26: #error "STANDALONE_WASM does not support shared memories yet".
+        # Disabling the pthreads avoids the inclusion of the library_pthread.js.
+        "-sUSE_PTHREADS=0",
+    ],
+)
+
+proxy_wasm_tests(
+    name = "tests",
+    plugins = [
+        ":plugin_cpp.wasm",
+    ],
+    tests = ":tests.textpb",
+)
--- a/plugins/samples/hmac_authcookie/plugin.cc
+++ b/plugins/samples/hmac_authcookie/plugin.cc
@@ -1,0 +1,93 @@
+// Copyright 2024 Google LLC
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+// [START serviceextensions_plugin_hmac_authcookie]
+#include <openssl/hmac.h>
+
+#include <iomanip>
+#include <sstream>
+#include <string>
+
+#include "absl/strings/str_split.h"
+#include "proxy_wasm_intrinsics.h"
+
+// Replace with your desired secret key.
+const std::string kSecretKey = "your_secret_key";
+
+class MyHttpContext : public Context {
+ public:
+  explicit MyHttpContext(uint32_t id, RootContext* root) : Context(id, root) {}
+
+  FilterHeadersStatus onRequestHeaders(uint32_t headers,
+                                       bool end_of_stream) override {
+    const auto token = getTokenFromCookie();
+    if (!token.has_value()) {
+      LOG_INFO("Access forbidden - missing HMAC cookie.");
+      sendLocalResponse(403, "", "Access forbidden - missing HMAC cookie.\n",
+                        {});
+      return FilterHeadersStatus::StopAllIterationAndWatermark;
+    }
+
+    const auto path = getRequestHeader(":path")->toString();
+    if (computeHmacSignature(path) != token.value()) {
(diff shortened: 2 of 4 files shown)


plugins/samples/hmac_authcookie/BUILD
```diff
@@ -0,0 +1,27 @@
...
+    deps = [
+        "@proxy_wasm_cpp_sdk//contrib:contrib_lib",
+        "@com_google_absl//absl/strings",
+        "@boringssl//:ssl",
+    ],
+    linkopts = [
+        # To avoid the error:
+        # library_pthread.js:26: #error "STANDALONE_WASM does not support shared memories yet".
+        # Disabling the pthreads avoids the inclusion of the library_pthread.js.
+        "-sUSE_PTHREADS=0",
```
martijneken, line 17: FWIW this will be fixed soon-ish in https://github.com/proxy-wasm/proxy-wasm-cpp-sdk/pull/173 by setting threads = "off" in the Bazel transition.

mpwarres: @walves-cit can you remove the `linkopts` setting (per @martijneken 's comment above)? Also I think it can be removed from plugins/samples/jwt_auth/BUILD as well (in a separate PR)

plugins/samples/hmac_authcookie/plugin.cc
```diff
@@ -0,0 +1,92 @@
...
+// Replace with your desired secret key.
+const std::string kSecretKey = "your_secret_key";
+
+class MyHttpContext : public Context {
+ public:
+  explicit MyHttpContext(uint32_t id, RootContext* root) : Context(id, root) {}
+
+  FilterHeadersStatus onRequestHeaders(uint32_t headers,
+                                       bool end_of_stream) override {
+    const auto token = getTokenFromCookie();
```
mpwarres, line 34: Expand type

plugins/samples/hmac_authcookie/plugin.cc
```diff
@@ -0,0 +1,92 @@
...
+    if (!token.has_value()) {
+      LOG_INFO("Access forbidden - missing HMAC cookie.");
+      sendLocalResponse(403, "", "Access forbidden - missing HMAC cookie.\n",
+                        {});
+      return FilterHeadersStatus::ContinueAndEndStream;
+    }
+
+    const auto path = getRequestHeader(":path")->toString();
```
mpwarres, line 42: Using a cookie to hold a HMAC for a specific path seems unrealistic to me: for that case, you'd typically include the HMAC in the path, like in #114 . The main benefit / use case of cookie-based HMAC would be to give a client a (usually time-bounded) token that can be used across multiple URLs.

For this example, I'd propose using a simple example cookie that contains the values (client IP, expiration timestamp, hash(client IP, expiration timestamp, secret)). More concretely, let `payload` = the string `client_ip + "," + expiration_timestamp`, and then the cookie would be `base64(payload) + "." + base64(computeHmacSignature(payload)).

You can determine the client IP by parsing the X-Forwarded-For header, whose format is documented at https://cloud.google.com/load-balancing/docs/https#x-forwarded-for_header. If that header isn't present or fails parsing, reject the request.

Then you can verify a request by:
1. obtaining client IP (reject if not present)
2. obtaining cookie (reject if not present)
3. Verify that cookie hash matches cookie payload (reject if no match)
4. get current time by calling `getCurrentTimeNanoseconds()`
5. Ensure that client IP matches cookie payload IP, and current time is earlier than cookie payload expiration time.

plugins/samples/hmac_authcookie/plugin.cc
```diff
@@ -0,0 +1,92 @@
...
+      if (cookie.first == "Authorization") {
+        return cookie.second;
+      }
+    }
+
+    return std::nullopt;
+  }
+
+  // Helper function to convert binary data to a hexadecimal string.
+  std::string toHexString(const unsigned char* data, size_t length) {
```
mpwarres, line 70: Suggest using absl::BytesToHexString

If that can't be used, suggest passing `data` and `length` as a `string_view`.

plugins/samples/hmac_authcookie/BUILD
walves-cit: Did not work. 😢 
I've got the following error:
```
wasm-ld: error: --shared-memory is disallowed by lto.tmp because it was not compiled with 'atomics' or 'bulk-memory' features.
```

plugins/samples/hmac_authcookie/plugin.cc
walves-cit: Done.... pls, take a look.
A lot of changes done 😅

plugins/samples/hmac_authcookie/BUILD
mpwarres: Ok, let's deal with this separately from this PR (which makes sense anyways, since it's present in a few other plugin samples as well)

plugins/samples/hmac_authcookie/plugin.cc
```diff
@@ -15,45 +15,125 @@
...
       LOG_INFO("Access forbidden - missing HMAC cookie.");
       sendLocalResponse(403, "", "Access forbidden - missing HMAC cookie.\n",
                         {});
       return FilterHeadersStatus::ContinueAndEndStream;
     }
 
-    const auto path = getRequestHeader(":path")->toString();
-    if (computeHmacSignature(path) != token.value()) {
+    const std::optional<std::pair<std::string, std::string>> payload_and_hash =
+        parseAuthorizationCookie(token.value());
```
mpwarres, line 77: optional: can dereference using `*token` instead of `token.value()`, since `has_value()` has been checked earlier. `*token` is shorter and doesn't have behavior that depends on build mode (i.e. whether or not to throw an exception)

plugins/samples/hmac_authcookie/plugin.cc
```diff
@@ -15,45 +15,125 @@
...
 class MyHttpContext : public Context {
  public:
-  explicit MyHttpContext(uint32_t id, RootContext* root) : Context(id, root) {}
+  explicit MyHttpContext(uint32_t id, RootContext* root)
+      : Context(id, root), root_(static_cast<MyRootContext*>(root)) {}
 
   FilterHeadersStatus onRequestHeaders(uint32_t headers,
                                        bool end_of_stream) override {
-    const auto token = getTokenFromCookie();
+    const std::optional<std::string> ip = getClientIp();
```
mpwarres, line 61: Nice structuring of this method! The factoring of checks/parsing into helper methods make it very clear.

plugins/samples/hmac_authcookie/plugin.cc
mpwarres: Very nice!

mpwarres (approved):

PR #118, opened 2024-09-12
feat: create a C++ sample plugin for creating a new cookie on HTTP response.

walves-cit:
This plugin is a create HTTP cookie showcase.

Technically, this is performed by validating if request was successful and case yes a new cookie is created on response.


plugins/samples/set_cookie/plugin.cc
```diff
@@ -0,0 +1,42 @@
...
+#include "absl/strings/numbers.h"
+#include "proxy_wasm_intrinsics.h"
+
+class MyHttpContext : public Context {
+ public:
+  explicit MyHttpContext(uint32_t id, RootContext* root) : Context(id, root) {}
+
+  FilterHeadersStatus onResponseHeaders(uint32_t headers,
+                                        bool end_of_stream) override {
+    const auto response_status = getResponseHeader(":status");
```
leonm1, line 25: Can you replace the auto with the underlying type, `WasmDataPtr`? The [Google C++ Style Guide](https://google.github.io/styleguide/cppguide.html#Type_deduction) indicates `auto` should be used `only if it makes the code clearer to readers who aren't familiar with the project, or if it makes the code safer`.

In this case, I think a method named "getResponseHeader" would return a `std::string`, but it instead returns a `WasmDataPtr`, so `auto` doesn't make it clearer.

leonm1 (changes requested):
Approved after requested changes.

plugins/samples/set_cookie/plugin.cc
walves-cit: Done

leonm1 (approved):
Thank you!

plugins/samples/set_cookie/plugin.cc
```diff
@@ -0,0 +1,42 @@
...
+  FilterHeadersStatus onResponseHeaders(uint32_t headers,
+                                        bool end_of_stream) override {
+    const WasmDataPtr response_status = getResponseHeader(":status");
+    int response_code;
+    // Check if response status is part of the success range (2XX).
+    if (response_status &&
+        absl::SimpleAtoi(response_status->view(), &response_code) &&
+        (response_code / 100 == 2)) {
+      // Change the new cookie according to your needs.
+      addResponseHeader("Set-Cookie",
```
mpwarres, line 32: While technically this example as it currently stands does set a cookie, I feel that it is somewhat incomplete as it does not look in the request for presence of an existing cookie, or embed / extract info from the cookie.

A more realistic example would be to use the cookie to preserve a session ID across multiple requests from the same user. So the flow would be roughly:

```
onRequestHeaders:
  check for presence of my_cookie in `Cookie` header
  if  my_cookie is present:
    extract/parse session ID value from it, and save in a field of MyHttpContext (e.g. MyHttpContext::session_id_). 
  else:
    leave MyHttpContext::session_id_ unset

onResponseHeaders:
  if MyHttpContext::session_id_ is set:
    log a message saying that current request is for existing session ID, including the session Id value in the log message
  else:
    generate new session ID
    log a message indicating that new session ID was created
    add "SetCookie" response header with new session ID
```

One thing to note from the above is that we only add `Set-Cookie` in the case where the request did not provide a cookie.

mpwarres (changes requested):

plugins/samples/set_cookie/plugin.cc
walves-cit: Makes sense but I have one question...
Given that we can have many requests it would be good to storage theirs session Ids in a unique map shared between the requests and maybe to have a expiration policy to avoid an endless memory allocation for each new user.... WDYT?

mpwarres: @walves-cit it should not be necessary to maintain a map of session IDs in the plugin. The way that cookies work is that the `Set-Cookie` response header instructs the client-side user agent (e.g. a browser) to store the cookie value on its side. It then echoes back the cookie value in subsequent requests. This effectively offloads storage of the cookie data (in this example, a session ID) to the client, so that the server (here, the plugin) doesn't need to maintain state.

So the steps outlined in my previous comment should be sufficient for managing cookies: if the client request doesn't contain a cookie, generate a new session ID and set it in the response using `Set-Cookie`; if the client request *does* contain a cookie, then extract the session ID from it.

WRT expiration policy of session IDs, cookie expiration times can be used for that: http://go/mdn/HTTP/Cookies#removal_defining_the_lifetime_of_a_cookie

For this example, in the case where the plugin is using `Set-Cookie` to set a new cookie, you could have it include a `Max-Age` setting to bound the cookie lifetime.

walves-cit: Understood now.... I did not know that the same plugin instance lives during all the request stream.
It was not making sense to have the `MyHttpContext::session_id_` if it could be shared between the requests. But not its fine I reread the docs and understood the lifecycle of the `RootContext` and the `Contexts`.
Tks.

walves-cit: Done... pls take a look

plugins/samples/set_cookie/plugin.cc
```diff
@@ -0,0 +1,81 @@
...
+    return FilterHeadersStatus::Continue;
+  }
+
+ private:
+  std::optional<std::string> session_id_;
+
+  // Try to get the session ID from the Cookie header.
+  static std::optional<std::string> getSessionIdFromCookie() {
+    const auto cookies = getRequestHeader("Cookie")->toString();
+    std::map<std::string, std::string> m;
```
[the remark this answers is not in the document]
walves-cit: Makes sense...Done

plugins/samples/set_cookie/plugin.cc
```diff
@@ -54,15 +54,14 @@ class MyHttpContext : public Context {
     const auto cookies = getRequestHeader("Cookie")->toString();
     std::map<std::string, std::string> m;
     for (absl::string_view sp : absl::StrSplit(cookies, "; ")) {
-      m.insert(absl::StrSplit(sp, absl::MaxSplits('=', 1)));
+      const std::pair<std::string, std::string> cookie =
+          absl::StrSplit(sp, absl::MaxSplits('=', 1));
+      if (cookie.first == "my_cookie") {
```
mpwarres, line 59: One more minor comment (sorry): since the cookie name is referenced in multiple places, can you define it as a constant?

plugins/samples/set_cookie/plugin.cc
```diff
@@ -0,0 +1,81 @@
...
+      return std::nullopt;
+    }
+
+    return token->second;
+  }
+
+  // Generate a unique random session ID.
+  static std::string generateSessionId() {
+    // Wasm VM does not support the random generation
+    // that involves a file system operation.
```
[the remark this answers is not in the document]
walves-cit: Exactly... I tried to use this Random lib and I got some issues and investigating the reason I found the following comments https://github.com/owasp-modsecurity/ModSecurity/pull/2763 and https://github.com/intel/modsecurity-wasm-filter/pull/16

mpwarres (approved):
LGTM, thanks!

plugins/samples/set_cookie/plugin.cc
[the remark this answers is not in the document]
martijneken: I have worked around this in cpp-sdk. This fix will be imported here: https://github.com/GoogleCloudPlatform/service-extensions/pull/124

If you use absl::random please be sure to put the BitGen in the root context -- it is somewhat expensive to initialize (a number of hostcalls).

PR #121, opened 2024-09-26
feat: feat: create a C++ sample plugin for A/B testing

walves-cit:
This plugin is a block request with particular URL/header showcase.

Technically, this is performed by the following hashing algorithm:

1. The user ID is hashed into an integer value between 0 and 99.
1. This integer value is compared to a predefined percentile (e.g., 50%).
1. If the hash value is less than or equal to the percentile, the request is served by v2 file. Otherwise, it is served by its original file.


plugins/samples/ab_testing/plugin.cc
```diff
@@ -0,0 +1,51 @@
...
+
+    // Checks if the current request is eligible to be served by file B.
+    //
+    // The decision is made by hashing the request path into an integer
+    // value between 0 and 99, then comparing the hash to a predefined
+    // percentile. If the hash value is less than or equal to the percentile,
+    // the request is served by file B. Otherwise, it is served by the original
+    // file.
+    if (absl::StartsWithIgnoreCase(path, a_path) &&
+        (std::hash<std::string_view>{}(path) % 100 <= percentile)) {
```
leonm1, line 40: Hashing the entire path can result in the user experience changing between reloads for a single user because the entire query string is included in the `:path` pseudo-header. (c/o @mpwarres, Envoy makes all request look like HTTP/2 requests, and the relevant section of the H2 spec is https://datatracker.ietf.org/doc/html/rfc9113#section-8.3.1).

For example, if this was used to A/B test a page with a table, a pagination token could mess this up:

Intent: redirect data/foo to data/foov2 for 50% of users:

Reality:

1. User navigates to `example.com/data/foo?page=1&user=1234`. Their user ID hashes to the experiment group, so their URL is rewritten to `example.com/data/foov2?page=1&user=1234`.
2. User goes to the second page, `example.com/data/foo?page=2&user=1234, but this time because of `?page=2`, the path is not hashed to the experiment group, so the URL is not rewritten.
3. The user receives a poor experience where the page experience is not "sticky".

(I am being pedantic because these are canonical examples which may be copied by many users, so the importance of defaults is important).

In this case, can you only hash the user ID? (also add a test where there are other query parameters that should not change the hash behavior)

walves-cit: I see... Makes sense for me.
Gonna change it following your recommendation.
Thanks for pointing this out.

plugins/samples/ab_testing/plugin.cc
```diff
@@ -0,0 +1,66 @@
...
+    const auto path = getRequestHeader(":path")->view();
+
+    // Checks if the current request is eligible to be served by v2 file.
+    //
+    // The decision is made by hashing the userID into an integer
+    // value between 0 and 99, then comparing the hash to a predefined
+    // percentile. If the hash value is less than or equal to the percentile,
+    // the request is served by the v2 file. Otherwise, it is served by the
+    // original file.
+    const auto usr = extractUserFromPath(path);
```
leonm1, line 42: Can you rename the variable to `user`? This goes against the [Google C++ Style Guide](https://github.com/GoogleCloudPlatform/service-extensions/pull/121/files#diff-f0449c7ae73157f77ebb17a0a75ff1f2dd7030e92fd68bb349ff05dc35d68dd3R42): "Do not abbreviate by deleting letters within a word"

leonm1 (changes requested):
Approved after making the requested change

plugins/samples/ab_testing/plugin.cc
walves-cit: Done

leonm1 (approved):
Thank you!

PR #129, opened 2024-11-13
feat: adding new c++ sample plugin to check for PII on responses

walves-cit:
This plugin is a PII on response check showcase.

Technically, this is performed by checking the response http headers, and the response body for the presence of credit card numbers. If found, the initial numbers will be masked.

### DIFF
--- a/plugins/README.md
+++ b/plugins/README.md
@@ -79,7 +79,7 @@ for your own plugin. Extend them to fit your particular use case.
 *   [Validate client JWT for authorization](samples/jwt_auth): Ensures user
     authentication by verifying an RS256-signed JWT token in the query string
     and subsequently removing it.
-    [Check for PII on response](samples/check_pii): Checks the response http headers,
+*   [Check for PII on response](samples/check_pii): Checks the response http headers,
     and the response body for the presence of credit card numbers. If found, the
     initial numbers will be masked.
 
--- a/plugins/README.md
+++ b/plugins/README.md
@@ -79,6 +79,9 @@ for your own plugin. Extend them to fit your particular use case.
 *   [Validate client JWT for authorization](samples/jwt_auth): Ensures user
     authentication by verifying an RS256-signed JWT token in the query string
     and subsequently removing it.
+    [Check for PII on response](samples/check_pii): Checks the response http headers,
+    and the response body for the presence of credit card numbers. If found, the
+    initial numbers will be masked.
 
 # Samples tests
 
--- a/plugins/README.md
+++ b/plugins/README.md
@@ -79,8 +79,8 @@ for your own plugin. Extend them to fit your particular use case.
 *   [Validate client JWT for authorization](samples/jwt_auth): Ensures user
     authentication by verifying an RS256-signed JWT token in the query string
     and subsequently removing it.
-*   [Check for PII on response](samples/check_pii): Checks the response http headers,
-    and the response body for the presence of credit card numbers. If found, the
+*   [Check for PII on response](samples/check_pii): Checks the response HTTP
+    headers and body for the presence of credit card numbers. If found, the
     initial numbers will be masked.
 
 # Samples tests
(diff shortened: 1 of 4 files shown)


plugins/README.md
```diff
@@ -79,6 +79,9 @@ for your own plugin. Extend them to fit your particular use case.
 *   [Validate client JWT for authorization](samples/jwt_auth): Ensures user
     authentication by verifying an RS256-signed JWT token in the query string
     and subsequently removing it.
+*   [Check for PII on response](samples/check_pii): Checks the response http headers,
```
mpwarres, line 82: nit: "response http headers, and the response body" -> "response HTTP headers and body"

plugins/samples/check_pii/plugin.cc
```diff
@@ -0,0 +1,84 @@
...
+#include "re2/re2.h"
+
+class MyRootContext : public RootContext {
+ public:
+  explicit MyRootContext(uint32_t id, std::string_view root_id)
+      : RootContext(id, root_id) {}
+
+  bool onConfigure(size_t) override {
+    // Credit card numbers in a 19 character hyphenated format.
+    // Compile the regex expression at plugin setup time.
```
mpwarres, line 26: append to comment: ", so that this expensive operation is only performed once, and not repeated with each request."

plugins/samples/check_pii/plugin.cc
```diff
@@ -0,0 +1,84 @@
...
+    card_match.emplace("\\d{4}-\\d{4}-\\d{4}-(\\d{4})");
+    return card_match->ok();
+  }
+
+  std::optional<re2::RE2> card_match;
+};
+
+// Checks the response http headers, and the response body for the presence of
+// credit card numbers. Mask the initial numbers case found.
```
mpwarres, line 35: Add: "Note that for illustrative purposes, this example is kept simple and does not handle the case of credit card numbers that are split across multiple onResponseBody() calls."

plugins/samples/check_pii/plugin.cc
```diff
@@ -0,0 +1,84 @@
...
+ public:
+  explicit MyHttpContext(uint32_t id, RootContext* root)
+      : Context(id, root), root_(static_cast<MyRootContext*>(root)) {}
+
+  FilterHeadersStatus onResponseHeaders(uint32_t headers,
+                                        bool end_of_stream) override {
+    const auto pii_header = getResponseHeader("google-run-pii-check");
+    if (pii_header && pii_header->view() == "true") {
+      check_body_ = true;
+      const auto result = getResponseHeaderPairs();
```
mpwarres, line 46: nit: I feel like it might help readability to expand some of these `auto` types--in the Google C++ style guide, the general guidance is to expand `auto` any place where the type of the variable is not readily apparent from the rest of the expression.

That said, a lot of our existing examples use `auto` fairly liberally, so maybe that argues for sticking with `auto` here for consistency, and doing a general pass to explicitly specify more types in a separate PR. So I'm not suggesting a change here, but just a heads up that I may send a future PR to spell out types more explicitly in this and other examples.

mpwarres (changes requested):
LGTM mod minor comments. Thanks!

plugins/samples/check_pii/plugin.cc
walves-cit: Agreed... So, I wont change here and wait for your PR to change it in more places. Tks

plugins/README.md
walves-cit: Done

plugins/samples/check_pii/plugin.cc
walves-cit: Done

plugins/samples/check_pii/plugin.cc
walves-cit: Done

mpwarres (approved):
LGTM, thanks!

PR #130, opened 2024-11-19
feat: Add body chunking support to test framework

dallen68:
feat: Add body chunking support to test framework

Small change:  Add option to use file as input and expectation for body invocations

- [ ] Tests pass


plugins/test/dynamic_test.cc
```diff
@@ -194,11 +239,32 @@ void DynamicTest::TestBody() {
     CheckPhaseResults("response_headers", invoke.result(), stream, res);
   }
   if (cfg_.response_body_size() > 0) {
-    for (auto& invoke : *cfg_.mutable_response_body()) {
-      auto res = stream.SendResponseBody(std::move(
-          *absl::WrapUnique(invoke.mutable_input()->release_content())));
-      ASSERT_VM_HEALTH("response_body", handle, stream);
-      CheckPhaseResults("response_body", invoke.result(), stream, res);
+    if (cfg_.chunking_plan_case() !=
```
mpwarres, line 242: This is very similar to the earlier logic from lines 205--230, can you factor it out into a helper method that can be called from both places?

plugins/test/dynamic_test.cc
```diff
@@ -287,12 +353,44 @@ void DynamicTest::BenchHttpHandlers(benchmark::State& state) {
...
+        auto body_chunk = ParseBodyInput(request_body.input());
+        BM_RETURN_IF_ERROR(body_chunk.status());
+        request_body_chunks.emplace_back(*body_chunk);
+      }
+    }
   }
   std::vector<std::string> response_body_chunks;
-  for (const auto& response_body : cfg_.response_body()) {
-    response_body_chunks.emplace_back(response_body.input().content());
+  if (cfg_.response_body_size() > 0) {
```
mpwarres, line 376: Can this be factored out into a helper method to reduce duplication with lines 356--373?

plugins/test/dynamic_test.cc
```diff
@@ -177,14 +199,47 @@ void DynamicTest::TestBody() {
...
+        body_chunking_plan !=
+            pb::Test::BodyChunkingPlanCase::BODY_CHUNKING_PLAN_NOT_SET) {
+      FAIL()
+          << "Cannot specify body_chunking_plan with multiple body invocations";
+      return;
     }
-  }
+    for (const pb::Invocation& invocation : invocations) {
+      TestHttpContext::Result body_result = TestHttpContext::Result{};
+      auto complete_input_body = ParseBodyInput(invocation.input());
```
mpwarres, line 218: suggest expanding this `auto` to the actual type, which isn't obvious from the statement by itself.

(See https://engdoc.corp.google.com/eng/doc/devguide/cpp/styleguide.md?cl=head#Type_deduction and go/totw/232 for general recommendations around when to use `auto`)

plugins/test/dynamic_test.cc
```diff
@@ -177,14 +199,47 @@ void DynamicTest::TestBody() {
...
+
+      std::vector<std::string> chunks;
+      if (body_chunking_plan !=
+          pb::Test::BodyChunkingPlanCase::BODY_CHUNKING_PLAN_NOT_SET) {
+        chunks = ChunkBody(*complete_input_body, cfg_);
+      } else {
+        chunks = {*complete_input_body};
+      }
+      for (std::string& body_chunk : chunks) {
+        auto res = invoke_wasm(std::move(body_chunk));
```
mpwarres, line 231: Expand `auto` to actual type

plugins/test/dynamic_test.cc
```diff
@@ -286,23 +337,23 @@ void DynamicTest::BenchHttpHandlers(benchmark::State& state) {
...
   }
-  std::vector<std::string> request_body_chunks;
-  for (const auto& request_body : cfg_.request_body()) {
-    request_body_chunks.emplace_back(request_body.input().content());
-  }
-  std::vector<std::string> response_body_chunks;
-  for (const auto& response_body : cfg_.response_body()) {
-    response_body_chunks.emplace_back(response_body.input().content());
-  }
+  auto request_body_chunks =
```
mpwarres, line 340: expand `auto` to actual type

plugins/test/dynamic_test.cc
```diff
@@ -286,23 +337,23 @@ void DynamicTest::BenchHttpHandlers(benchmark::State& state) {
...
+      PrepBodyCallbackBenchmark(cfg_, cfg_.request_body());
+  BM_RETURN_IF_ERROR(request_body_chunks.status());
+  auto response_body_chunks =
```
mpwarres, line 343: expand `auto` to actual type

plugins/test/dynamic_test.cc
```diff
@@ -521,15 +572,56 @@ absl::StatusOr<TestHttpContext::Headers> DynamicTest::ParseHeaders(
...
+  auto body_chunking_plan = test.body_chunking_plan_case();
+  if (invocations.size() != 1 &&
+      body_chunking_plan !=
+          pb::Test::BodyChunkingPlanCase::BODY_CHUNKING_PLAN_NOT_SET) {
+    return absl::InvalidArgumentError(
+        "Cannot specify body_chunking_plan with multiple body invocations");
+  }
+  std::vector<std::string> chunks;
+  for (const pb::Invocation& invocation : invocations) {
+    auto complete_input_body = ParseBodyInput(invocation.input());
```
mpwarres, line 613: expand `auto` to actual type

plugins/test/runner.proto
```diff
@@ -49,7 +51,9 @@ message Input {
   bytes content = 2;
 
   // Serialized HTTP content, but provided via a separate file.
-  // File path can be absolute or relative to the test config file.
+  // For body input, don't include any content other than the body you would
```
mpwarres, line 54: Doesn't this comment apply to the `content` field as well? If so, consider moving it to the top-level doc comment for the `Input` message. Or just remove it, I would have assumed that the file only contains body content, when used for body input.

mpwarres (commented):
LGTM mod remaining comments, all trivial

mpwarres:
Also, there appear to be some failures in the ASan and TSan runs, those need to be resolved before this can be merged.

plugins/test/runner.proto
dallen68: This is to differentiate from headers, where the file input is more than just the header. The file input includes the entire request.

mpwarres: ok, SG

mpwarres (approved):

PR #131, opened 2024-11-27
fix:Check for incorrect return-type during plugin compilation

dallen68:
fix: Check for incorrect return-type during plugin compilation

- [ ] Tests pass

### DIFF
--- a/plugins/plugins.bzl
+++ b/plugins/plugins.bzl
@@ -29,8 +29,9 @@ def proxy_wasm_plugin_rust(**kwargs):
         **kwargs
     )
 
-def proxy_wasm_plugin_cpp(**kwargs):
+def proxy_wasm_plugin_cpp(copts = [], **kwargs):
     proxy_wasm_cc_binary(
+        copts = copts + ["-Werror=return-type"],
         **kwargs
     )
 


plugins/plugins.bzl
```diff
@@ -29,8 +29,9 @@ def proxy_wasm_plugin_rust(**kwargs):
         **kwargs
     )
 
-def proxy_wasm_plugin_cpp(**kwargs):
+def proxy_wasm_plugin_cpp(copts = [], **kwargs):
     proxy_wasm_cc_binary(
+        copts = copts + ["-Werror=return-type"],
```
mpwarres, line 34: Any reason not to do "-Werror" by itself, which would turn all warnings into errors?

mpwarres (approved):
LGTM mod comment. Also, I think that adding a space after "fix:" in the PR title may satisfy the "Conventional Commits" linter

plugins/plugins.bzl
dallen68: There are build relating warnings tied to other dependencies that we don't necessarily want to create an error for. I think "-Werror" is too heavy handed.

mpwarres: SGTM, thanks!

martijneken, line 34: What was the reasoning for adding this here instead of proxy-wasm-cpp-sdk?
https://github.com/proxy-wasm/proxy-wasm-cpp-sdk/blob/a982ad089d1962f1b92f4343c910c0be6b6e6280/bazel/defs.bzl#L73

I suspect that if we found this sharp edge in samples, then end users of the SDK could run into this as well.

mpwarres: The `proxy_wasm_cc_binary` rule currently [provided](https://github.com/proxy-wasm/proxy-wasm-cpp-sdk/blob/a982ad089d1962f1b92f4343c910c0be6b6e6280/bazel/defs.bzl#L73) by proxy-wasm-cpp-sdk leaves it up to the caller of that rule to specify the `copts` (or other options) it wants, much like `cc_binary`. I wouldn't really expect `cc_binary` to choose `copts` for me, I was thinking the same argument could be made for `proxy_wasm_cc_binary`.

Maybe this is a strained example, but suppose a plugin depends on some third-party library that fails compilation with `-Werror=return-type`. In that case, a user would have to modify SDK build rules to get their plugin to compile. Granted, this is probably fairly unlikely, so I could see the argument that adding at least `-Werror=return-type` is a worthwhile tradeoff. Adding blanket `-Werror` would be too strong IMO, though.

PR #135, opened 2024-12-05
feat: create a Rust sample plugin for creating a new cookie on HTTP response.

pweiber:
Checks for an existing my_cookie in incoming HTTP requests:

If the cookie exists, it logs the current session ID.
If absent, it generates a new random session ID, sets it in the Set-Cookie header, and logs the creation of the new session.

This example contains only a Rust version.

### DIFF
--- a/plugins/samples/set_cookie/plugin.rs
+++ b/plugins/samples/set_cookie/plugin.rs
@@ -1,0 +1,112 @@
+// Copyright 2024 Google LLC
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+// [START serviceextensions_plugin_set_cookie]
+use proxy_wasm::traits::*;
+use proxy_wasm::types::*;
+use rand::Rng;
+
+// Define the cookie name as a constant
+const COOKIE_NAME: &str = "my_cookie";
+
+proxy_wasm::main! {{
+    proxy_wasm::set_log_level(LogLevel::Info);
+    proxy_wasm::set_root_context(|_| -> Box<dyn RootContext> {
+        Box::new(MyRootContext)
+    });
+}}
+
+struct MyRootContext;
+
+impl Context for MyRootContext {}
+
+impl RootContext for MyRootContext {
+
+    fn create_http_context(&self, _context_id: u32) -> Option<Box<dyn HttpContext>> {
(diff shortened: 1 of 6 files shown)


plugins/samples/set_cookie/plugin.rs
```diff
@@ -0,0 +1,119 @@
...
+// limitations under the License.
+
+// [START serviceextensions_plugin_set_cookie]
+use proxy_wasm::traits::*;
+use proxy_wasm::types::*;
+use rand::rngs::SmallRng;
+use rand::{RngCore, SeedableRng};
+
+// Define the cookie name as a constant
+const COOKIE_NAME: &str = "my_cookie";
```
mpwarres, line 22: should this be `static COOKIE_NAME: ...`?

plugins/samples/set_cookie/plugin.rs
```diff
@@ -0,0 +1,119 @@
...
+    }
+
+    /// Generates a random `u32` session ID.
+    fn generate_random_session_id(&self) -> u32 {
+        let seed = std::time::SystemTime::now()
+            .duration_since(std::time::UNIX_EPOCH)
+            .unwrap_or_default()
+            .as_nanos() as u64;
+
+        let mut rng = SmallRng::seed_from_u64(seed);
```
mpwarres, line 115: Suggest instantiating and seeding the rng in RootContext, and then on_http_response_headers() only needs to call `rng.next_u32()`

plugins/samples/set_cookie/plugin.rs
pweiber: I believe that in this scenario we can keep the const, If I'm not wrong since we are not changing it the fixed string for the const is more efficient as the compiler can inline it

plugins/samples/set_cookie/plugin.rs
pweiber: Changed to the RootContext, let me know if it looks fine now

plugins/samples/set_cookie/plugin.rs
mpwarres: SG, thanks

plugins/samples/set_cookie/plugin.rs
```diff
@@ -0,0 +1,123 @@
...
+proxy_wasm::main! {{
+    proxy_wasm::set_log_level(LogLevel::Info);
+    proxy_wasm::set_root_context(|_| -> Box<dyn RootContext> {
+        let seed = std::time::SystemTime::now()
+            .duration_since(std::time::UNIX_EPOCH)
+            .unwrap_or_default()
+            .as_nanos() as u64;
+
+        Box::new(MyRootContext {
+            rng: SmallRng::seed_from_u64(seed),
```
mpwarres, line 33: I think you need to make this a std::rc::Rc<SmallRng> and use the same shared SmallRng instance from MyHttpContext. Otherwise, this is repeatedly clone'ing SmallRng with a given seed, so each MyHttpContext will see the same "random" value from its calls to next_u32().

mpwarres (commented):
LGTM once rng is wrapped in an Rc so that all MyHttpContexts use the same instance (and therefore get distinct random numbers)

PR #139, opened 2024-12-10
feat: create a Rust sample plugin for overwriting error codes from 5XX to 4XX classes.

pweiber:
Rust version for the overwriting error codes.

This is performed by replacing the 5XX error code response to 4XX.


plugins/samples/overwrite_errcode/plugin.rs
```diff
@@ -0,0 +1,51 @@
...
+                    self.set_http_response_header(":status", Some(&new_status.to_string()));
+                }
+            }
+        }
+        return Action::Continue;
+    }
+}
+
+// Function that remaps all 5xx responses to 404.
+fn map_response_code(response_code: i32) -> i32 {
```
dallen68, line 43: I don't think we need this helper function. You could just do `self.set_http_response_header(":status", "404");`

dallen68 (changes requested):

dallen68 (approved):

PR #140, opened 2024-12-12
feat: create a Rust sample plugin for A/B testing

pweiber:
Rust version for the A/B testing.

If the hash value is less than or equal to the percentile, the request is served by v2 file. Otherwise, it is served by its original file.


plugins/samples/ab_testing/plugin.rs
```diff
@@ -0,0 +1,72 @@
...
+        }
+    }
+    "".to_string()
+}
+
+// Compute hash of `user` and determine if it falls into the given percentile
+fn should_use_v2(user: &str) -> bool {
+    // Calculate a deterministic hash from user.
+    // Sum the ASCII values of the characters and then offset by length.
+    let sum: u64 = user.bytes().map(u64::from).sum();
```
dallen68, line 66: Is there a reason why we aren't using the default hashing implementation for slices in Rust? https://doc.rust-lang.org/stable/std/hash/trait.Hash.html#method.hash_slice

dallen68 (changes requested):

plugins/samples/ab_testing/plugin.rs
pweiber: I did this way because the tests were based on the c++ code, for one of them if I used the default hashing I was getting different values

for the "WithHashAboveThresholdKeepTheSame"

C++ computes std::hash("userAAA") % 100 = 63 (63 > 50 → no redirect).
Rust’s DefaultHasher produces a different value (e.g., hash % 100 ≤ 50 → redirect).

Using like this I'm getting
// For "userAAA": 642 + 21 = 663 → 63%100

Let me know if it is fine to keep it like this or if we should adjust for both

dallen68: Sounds good. Maybe add a comment saying why were are hashing by hand instead of using the rust standard library.

dallen68 (approved):

pweiber:
@dallen68 Let me know if the comment is fine like that or if we should change something

PR #151, opened 2025-01-07
feat: Add a Github Action to publish callouts samples

martijneken:
Changes:
- publish 4 sample images (can add later as needed)
- move basic example into its own directory (update docs to match)
- use launcher.gcr.io to avoid GCR auth issues
- standardize on secret name GAR_TOKEN
- standardize on apt-get over apt

Check that:
- [x] Tests pass


mpwarres (approved):

jstraceski:
Looks good to me! But we should probably also update the sphinx documentation to reflect the basic callout service changes and readme changes:
i.e.
docs/python/extproc/basic_callout_server.rst
docs/python/readme

martijneken:
> Looks good to me! But we should probably also update the sphinx documentation to reflect the basic callout service changes and readme changes: i.e. docs/python/extproc/basic_callout_server.rst docs/python/readme

Thanks, I've updated the non-generated docs

PR #152, opened 2025-01-07
feat: Add initial sample for go plugin

leonm1:


plugins/samples/check_pii/plugin.go
```diff
@@ -0,0 +1,101 @@
...
+		if err != nil {
+			proxywasm.SendHttpResponse(500, [][2]string{}, []byte(fmt.Sprintf("%v", err)), 0)
+		}
+	}()
+	bytes, err := proxywasm.GetHttpResponseBody(0, numBytes)
+	if err != nil {
+		panic(err)
+	}
+	bytes = ctx.creditCardRegex.ReplaceAll(bytes, []byte("XXXX-XXXX-XXXX-${1}"))
+	err = proxywasm.ReplaceHttpResponseBody(bytes)
```
martijneken, line 94: Avoiding this call when no substitutions are made would be slightly more efficient. You could avoid the string copy and the hostcall in most cases. You could probably get there with https://pkg.go.dev/regexp#Regexp.FindAllIndex

And if you can call ReplaceHttpResponseBody for substrings you could do targeted replacements of CC strings.

plugins/samples/check_pii/plugin.go
```diff
@@ -0,0 +1,101 @@
...
+type httpContext struct {
+	types.DefaultHttpContext
+	creditCardRegex *regexp.Regexp
+	checkBody       bool
+}
+
+func (*context) NewPluginContext(contextID uint32) types.PluginContext {
+	return &context{creditCardRegex: regexp.MustCompile("\\d{4}-\\d{4}-\\d{4}-(\\d{4})")}
+}
+func (pluginContext *context) NewHttpContext(contextID uint32) types.HttpContext {
```
martijneken, line 45: Please add this caveat somewhere: https://github.com/GoogleCloudPlatform/service-extensions/blob/6170c6b99aac4fc4cd306dc58544eaf173b44ca4/plugins/samples/check_pii/plugin.cc#L37-L39

This code does not handle PII that spans chunk boundaries.

martijneken (changes requested):
Exciting!

leonm1:
I removed the `//...` expression from ci, since it failed to build the gazelle target under asan/ubsan. I'd rather not remove the gazelle target, since it's the primary way to consume Go modules with bazel.

plugins/samples/check_pii/plugin.go
```diff
@@ -0,0 +1,101 @@
...
+	"github.com/proxy-wasm/proxy-wasm-go-sdk/proxywasm"
+	"github.com/proxy-wasm/proxy-wasm-go-sdk/proxywasm/types"
+)
+
+func main() {}
+func init() {
+	proxywasm.SetVMContext(&context{})
+}
+
+type context struct {
```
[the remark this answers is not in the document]
martijneken: Thanks! Filed https://github.com/proxy-wasm/proxy-wasm-go-sdk/issues/5 for better usability.

plugins/samples/check_pii/plugin.go
martijneken: Thanks for the deep dive.

I agree a Cord would be a better internal representation for the host! Future work.

martijneken (approved):

PR #156, opened 2025-01-14
feat: create a Rust check for PII responses sample plugin

pweiber:
Rust version for the check pii c++ version in #136 

The tests used are from #136

Consider merging it before this one

Includes:

- Mask for CC numbers in 16-digit hyphenated format.
- Mask for a 10 digit numeric code (In addition to the already CC check)

### DIFF
--- a/plugins/samples/check_pii/plugin.rs
+++ b/plugins/samples/check_pii/plugin.rs
@@ -1,4 +1,4 @@
-// Copyright 2025 Google LLC
+// Copyright 2026 Google LLC
 //
 // Licensed under the Apache License, Version 2.0 (the "License");
 // you may not use this file except in compliance with the License.


plugins/samples/check_pii/plugin.rs
```diff
@@ -0,0 +1,169 @@
...
+proxy_wasm::main! {{
+    proxy_wasm::set_log_level(LogLevel::Trace);
+    proxy_wasm::set_root_context(|_| -> Box<dyn RootContext> {
+        Box::new(MyRootContext::new())
+    });
+}}
+
+struct MyRootContext {
+    card_matcher: Option<Arc<Regex>>,
+    code10_matcher: Option<Arc<Regex>>,
```
dallen68, line 31: Can we use Rc instead of Arc here? Since we aren't interacting with multiple threads we don't need to use Arc.

dallen68 (changes requested):

dallen68 (changes requested):
Tests fail

pweiber:
> Tests fail

@dallen68 It is because I'm using the #136 test file, I can push it here but I just realized that I might need to update the GO code  to reflect the c++ changes as well otherwise it will fail again.

Is that ok if I include the GO changes here? Then it should be fine to be merged after that

dallen68:
We can wait on the other PR to merge to avoid adding GO changes.

PR #158, opened 2025-01-15
chore: Use vendored mode for more efficient Rust plugin deps

### LINKED ISSUE #154: Use vendored mode for more efficient Rust plugin deps
Use crate_universe vendored mode so that non-Rust builds don't need to build all Rust deps.

Example from cpp-host: https://github.com/proxy-wasm/proxy-wasm-cpp-host/tree/main/bazel/cargo/wasmtime

We could split into multiple Cargo.toml files / directories, but that seems like unnecessary work that could bloat the samples.

dallen68:
Fixes #154


plugins/bazel/cargo/Cargo.toml
```diff
@@ -3,13 +3,13 @@ name = "service-extensions"
...
+# Reference: //bazel/cargo/remote:proxy-wasm (etc)
 [dependencies]
-proxy-wasm = "0.2.1"
-log = "0.4"
-regex = "1.9"
-url = "2.4"
+proxy-wasm = "=0.2.1"
+log = "=0.4"
+regex = "=1.9"
+url = "=2.4"
```
martijneken, line 12: Can we roll back the addition of the "=" prefixes, or is it needed for some reason?

martijneken:
header-check expects LICENSE headers in these generated files:
```
plugins/bazel/cargo/remote/alias_rules.bzl is missing a valid license header.
plugins/bazel/cargo/remote/crates.bzl is missing a valid license header.
plugins/bazel/cargo/remote/defs.bzl is missing a valid license header.
```

Just need to update the ignoreFiles pattern here:
https://github.com/GoogleCloudPlatform/service-extensions/blob/main/.github/header-checker-lint.yml

plugins/bazel/cargo/Cargo.toml
dallen68: Cargo will sometimes use the newer version of a crate if there is one available. The "=" prefix tells cargo to only use the version specified.

.github/header-checker-lint.yml
```diff
@@ -30,3 +30,4 @@ sourceFileExtensions:
 
 ignoreFiles:
   - "cargo/**"  # Ignore BUILD rules generated by cargo-raze
```
martijneken, line 32: Please remove this stale line, we no longer use cargo-raze

plugins/bazel/cargo/Cargo.toml
martijneken: Is repinning to later patch versions a bad thing? The last known working dependency tree is captured in the generated build files, so unless you regen those with --repin there is no change. I would prefer not "hard pinning" if possible.

dallen68: I've already ran into some cases where repining the latest version has resulted in plugin samples breaking. Also, after repining there isn't an easy way to rollback when a breakage would occur. I think it's better to "hard pin" these top level dependencies, otherwise when adding a new dependency and repining you could unintentionally break other plugins since they now have "updated" dependencies. I think if we weren't Wasm based it would make more sense to use the latest version of libraries, but because we are we want to avoid pulling in changes that fundamentally don't work for us(threading, IO, etc.). Rust/Cargo will still recognize a crate as stable or healthy when these changes occur and pull in changes that silently break sample plugins.

mpwarres: > I've already ran into some cases where repining the latest version has resulted in plugin samples breaking

Can you elaborate on what sort of breakages you saw? If our samples break with patch version updates to dependencies, I'm wondering if there might be an underlying issue we need to fix. Maybe I'm missing something, but ISTM that fixing this by pinning to exact versions is addressing a symptom rather than the cause.

dallen68: After an offline discussion, I agree that we should handle breakages as they occur and only pin dependencies as needed as a temporary solution.

martijneken:
Please sync and fix CI

martijneken (approved):

PR #168, opened 2025-01-24
feat: create a Go sample plugin to create a new cookie on HTTP response

pweiber:
Go version for the set cookie sample.


plugins/samples/set_cookie/plugin.go
```diff
@@ -0,0 +1,124 @@
...
+
+	if ctx.sessionID != nil {
+		proxywasm.LogInfof("This current request is for the existing session ID: %s", *ctx.sessionID)
+	} else {
+		// Generate new session ID
+		newSessionID := fmt.Sprintf("%d", generateRandom())
+		proxywasm.LogInfof("New session ID created for the current request: %s", newSessionID)
+
+		// Set the cookie
+		cookieValue := fmt.Sprintf("%s=%s; Path=/; HttpOnly", cookieName, newSessionID)
```
leonm1: Please construct the cookie using "net/http" library's `Cookie` struct and use Cookie.String() instead of formatting it manually.

leonm1: This should now be available in service-extensions repo main branch

pweiber: Just re-run the tests and seems to be working fine now, thanks for the update on this!

PR #174, opened 2025-01-29
feat: create a Go sample for updating http body chunks

walves-cit:
This plugin is an HTTP body chunking update showcase.

Technically, this is performed by appending some strings onto the body chunking from request and response.

This examples contains only a Go version.

### DIFF
--- a/plugins/samples/body_chunking/plugin.go
+++ b/plugins/samples/body_chunking/plugin.go
@@ -1,0 +1,88 @@
+// Copyright 2025 Google LLC
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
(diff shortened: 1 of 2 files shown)


plugins/samples/body_chunking/plugin.go
```diff
@@ -0,0 +1,88 @@
...
+
+type httpContext struct {
+	types.DefaultHttpContext
+}
+
+func (*vmContext) NewPluginContext(contextID uint32) types.PluginContext {
+	return &pluginContext{}
+}
+
+func (*pluginContext) NewHttpContext(uint32) types.HttpContext {
```
leonm1: Since vmContext and pluginContext are not used, please use the SetHttpContext entrypoint in init() and remove the unneeded contexts.

PR #175, opened 2025-01-30
feat: create a Go sample for validate http header tokens based on plugin config file

walves-cit:
This plugin is an HTTP token header validation using config file showcase.

Technically, this is performed by validating if the HTTP User token is allowed based on a config file deny list.

This examples contains only a Go version.

### DIFF
--- a/plugins/samples/config_denylist/plugin.go
+++ b/plugins/samples/config_denylist/plugin.go
@@ -1,0 +1,92 @@
+// Copyright 2025 Google LLC
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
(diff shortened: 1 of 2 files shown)


plugins/samples/config_denylist/plugin.go
```diff
@@ -0,0 +1,92 @@
...
+}
+
+type pluginContext struct {
+	types.DefaultPluginContext
+	tokens map[string]struct{}
+}
+
+type httpContext struct {
+	types.DefaultHttpContext
+	pluginContext *pluginContext
```
leonm1: Can you make httpContext store a pointer to `tokens` instead? It'd be a lot simpler than calling methods on ctx.pluginContext.

PR #183, opened 2025-02-12
fix: prevent plugin crash when calling setBuffer with start>size of buffer

dallen68:


plugins/README.md
```diff
@@ -138,6 +138,8 @@ for your own plugin. Extend them to fit your particular use case.
     initial numbers will be masked.
 *   [Validate client token on query string using HMAC](samples/hmac_authtoken):
     Check the client request URL for a valid token signed using HMAC.
+*   [Modify body based on request headers](samples/configurable_body_modification):
+    Check for specific request headers that set parameters for corresponding setBuffer calls in body callback.
```
martijneken, line 142: I appreciate you testing that the fix to the test framework works, but IMO this configurable plugin doesn't really help Service Extensions users because it doesn't model a typical plugin capability.

I think it could be really useful as a unit test in the Google internal CppHost implementation, just not here in the samples repo. The framework fix itself LGTM.

martijneken (approved):

PR #184, opened 2025-02-13
fix: Remove benchmarks from sample tests

dallen68:
Benchmarks in tests caused ASan and TSan tests to occasionally fail due to timeouts.

### DIFF
--- a/plugins/samples/add_header/tests.textpb
+++ b/plugins/samples/add_header/tests.textpb
@@ -1,6 +1,5 @@
 test {
   name: "AddsRequestHeader"
-  benchmark: true
   request_headers {
     result {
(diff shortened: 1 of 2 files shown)


leonm1 (approved):

martijneken:
> Benchmarks in tests caused ASan and TSan tests to occasionally fail due to timeouts.

Good finding, but could we instead skip running benchmarks in CI? I think you can just add `--test_arg=--nobench` here:
https://github.com/GoogleCloudPlatform/service-extensions/blob/73e8ada9dee1de708568d9fa33cfd9734cb67da4/.github/workflows/plugins-test.yml#L82

PR #186, opened 2025-02-19
fix: Clarify relationship between immediate responses and callback flow

dallen68:


plugins/test/framework.cc
```diff
@@ -78,6 +78,7 @@ ContextOptions& TestContext::options() const {
 
 proxy_wasm::BufferInterface* TestHttpContext::getBuffer(
     proxy_wasm::WasmBufferType type) {
+  if (immediate_response_) return nullptr;
```
martijneken, line 81: I don't think these hostcall short-circuits match production behavior. Let's remove them.

plugins/test/framework.cc
```diff
@@ -185,20 +195,30 @@ TestHttpContext::Result TestHttpContext::SendRequestHeaders(
 TestHttpContext::Result TestHttpContext::SendRequestBody(std::string body) {
   phase_logs_.clear();
   result_ = Result{};
-  body_buffer_.setOwned(std::move(body));
   current_callback_ = TestHttpContext::CallbackType::RequestBody;
+  // Don't invoke callbacks if immediate response has been sent.
+  if (immediate_response_) return std::move(result_);
```
martijneken, line 200: These short circuits should be one-liners IMO so they don't cause new conditionals below:

`if (immediate_response_) return Result{};`

plugins/test/framework.cc
```diff
@@ -155,8 +163,9 @@ proxy_wasm::WasmResult TestHttpContext::sendLocalResponse(
     uint32_t response_code, std::string_view body_text,
     proxy_wasm::Pairs additional_headers, uint32_t grpc_status,
     std::string_view details) {
-  if (phase_ != proxy_wasm::WasmHeaderMapType::RequestHeaders &&
-      phase_ != proxy_wasm::WasmHeaderMapType::ResponseHeaders) {
+  if (current_callback_ != TestHttpContext::CallbackType::RequestHeaders &&
+      current_callback_ != TestHttpContext::CallbackType::RequestBody &&
```
martijneken, line 167: This might be a good change but I'm not sure. We need to check on the Envoy behavior of immediate response during request/response body processing in STREAMED and FULL_DUPLEX body modes. Let's follow up in person.

In any case, ImmediateResponse in a body handler should be a separate PR.

plugins/test/dynamic_test.cc
```diff
@@ -392,7 +402,66 @@ void DynamicTest::CheckSideEffects(const std::string& phase,
...
+      (expect.has_immediate() || expect.has_header_size() != 0 ||
+       expect.no_header_size() != 0 || expect.headers_size() != 0 ||
+       expect.body_size() != 0 || expect.log_size() != 0)) {
+    ADD_FAILURE() << absl::Substitute(
+        "[$0] Expected results after sending immediate response in previous "
+        "callback.",
+        phase);
+    return;
+  }
+  // Check immediate response.
```
martijneken, line 417: Please move this logic back down, I can't see the diff.

plugins/test/dynamic_test.cc
```diff
@@ -392,7 +402,66 @@ void DynamicTest::CheckSideEffects(const std::string& phase,
 void DynamicTest::CheckPhaseResults(const std::string& phase,
                                     const pb::Expectation& expect,
                                     const TestContext& context,
-                                    const TestHttpContext::Result& result) {
+                                    const TestHttpContext::Result& result,
+                                    bool* immediate_response_sent) {
+  if (*immediate_response_sent &&
```
martijneken, line 407: IMO tracking this boolean is unnecessarily complex. If you needed it, you could instead expose the new immediate_response_ boolean tracked inside TestHttpContext.

But my preferred approach is to not change anything in DynamicTest: the changes you made in TestHttpContext are sufficient to make any post-immediate-response tests fail, because we never invoke those callbacks anymore.

plugins/samples/immediate_response/tests.textpb
```diff
@@ -0,0 +1,65 @@
+test {
+  name: "ImmediateResponseFromRequestHeaders"
```
martijneken, line 2: Similar feedback to PR#183: this sample + test doesn't help Service Extensions users get stuff done.

We have samples that use ImmediateResponse in headers:
https://github.com/search?q=repo%3AGoogleCloudPlatform%2Fservice-extensions%20sendLocalResponse&type=code

We should perhaps add a body-based ImmediateResponse sample, if we determine that is a good thing (TBD, see other comment). But we should write a sample for a real use case, such as a WAF that inspects request body. ImmediateResponse in a body handler should be a separate PR.

martijneken (changes requested):

PR #187, opened 2025-02-24
fix: Prevent benchmarks in automated workflow

dallen68:


mpwarres (approved):
LGTM, suggest mentioning in the PR description that this replaces the change in #184 .

PR #189, opened 2025-03-02
feat: Add custom stats to benchmarking

martijneken:
Adds stats for memory / logging / binary size in bytes.

The memory stats could use further improvement since they
1. Don't account for stream parallelism
2. Don't give insight into wasm-side allocations

But this is a good enough starting point.

Also adds a flag to optionally disable unit tests.

- [x] Tests pass

### DIFF
--- a/plugins/test/dynamic_test.cc
+++ b/plugins/test/dynamic_test.cc
@@ -216,7 +216,8 @@ void DynamicTest::TestBody() {
     }
     for (const pb::Invocation& invocation : invocations) {
       TestHttpContext::Result body_result = TestHttpContext::Result{};
-      absl::StatusOr<std::string> complete_input_body = ParseBodyInput(invocation.input());
+      absl::StatusOr<std::string> complete_input_body =
+          ParseBodyInput(invocation.input());
       if (!complete_input_body.ok()) {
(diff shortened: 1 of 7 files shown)


plugins/test/dynamic_test.cc
```diff
@@ -278,17 +279,42 @@ void DynamicTest::TestBody() {
...
   auto handle = *load_wasm;
 
   // Benchmark plugin initialization and teardown.
+  bool first = true;
   for (auto _ : state) {
     // Create root context (start) and call configure on it.
     auto plugin_init = InitializePlugin(handle);
     BM_RETURN_IF_ERROR(plugin_init);
+
+    // Emit root stats before teardown. After teardown the context is gone.
```
leonm1, line 309: This means we won't accurately reflect any logs in the OnDone/OnDelete callbacks for the plugin or VM context. That doesn't seem ideal, but it could be passable behavior. Thoughts?

Alternative seems to be owning logging_bytes_ outside of the `TestContext`, possibly in `TestWasm`?

leonm1 (approved):
LGTM, mod suggested change to own logging_bytes_ in `TestWasm`.

PR #192, opened 2025-03-04
feat: Add content injection sample with benchmarks

dallen68:
Updates comments in html_domain_rewrite plugin.


plugins/samples/content_injection/plugin.rs
```diff
@@ -0,0 +1,180 @@
...
+use std::cell::RefCell;
+use std::error::Error;
+use std::rc::Rc;
+
+proxy_wasm::main! {{
+    proxy_wasm::set_log_level(LogLevel::Trace);
+    proxy_wasm::set_http_context(|_, _| -> Box<dyn HttpContext> { MyHttpContext::new()});
+}}
+
+struct MyOutputSink {
```
leonm1, line 29: Can you give these structs more descriptive names? Perhaps SharedByteSink and ContentInjectionHttpContext?

plugins/samples/content_injection/plugin.rs
```diff
@@ -0,0 +1,180 @@
...
+            .expect("Expected valid rewriter. Got None")
+            .end()?;
+        return Ok(());
+    }
+}
+
+impl<'a> Context for MyHttpContext<'a> {}
+
+impl<'a> HttpContext for MyHttpContext<'a> {
+    fn on_http_request_headers(&mut self, _: usize, _: bool) -> Action {
```
leonm1, line 122: IIUC, I think this should work without this function implemented at all

plugins/samples/content_injection/plugin.rs
```diff
@@ -0,0 +1,180 @@
...
+impl<'a> MyHttpContext<'a> {
+    pub fn new() -> Box<MyHttpContext<'a>> {
+        let output = Rc::new(RefCell::new(Vec::new()));
+        let completed = Rc::new(RefCell::new(false));
+        Box::new(MyHttpContext {
+            output: output.clone(),
+            completed: completed.clone(),
+            rewriter: Some(HtmlRewriter::new(
+                Settings {
+                    element_content_handlers: vec![element!("head", move |el| {
```
leonm1, line 82: Can you add a comment or extract this into a helper function that makes it clear that this is where a customer would add logic for what to inject?

It seems like it should be in the on_response_body handler, but it's actually in the constructor due to the design of the HTML Rewriter.

leonm1 (changes requested):

plugins/samples/content_injection/plugin.rs
dallen68: Done

plugins/samples/content_injection/plugin.rs
dallen68: Removed

plugins/samples/content_injection/plugin.rs
dallen68: Done

leonm1: Can you also rename MyHttpContext to something like ContentInjectionHttpContext?

leonm1 (approved):

PR #196, opened 2025-03-04
feat: Add reCAPTCHA challenge sample

dallen68:
- [ ] Tests pass

### DIFF
--- a/plugins/README.md
+++ b/plugins/README.md
@@ -142,6 +142,10 @@ for your own plugin. Extend them to fit your particular use case.
 *   [Rewrite domains in html response body](samples/html_domain_rewrite/): Parse
     html in response body chunks and replace insances of "foo.com" with
     "bar.com" in `<a href=***>`.
+*   [Enable reCAPTCHA challenge on response body](samples/enable_recaptcha/):
+    Enable reCAPTCHA challenge on response body, following guidence at
+    [Automatically bind the challenge to a button](https://developers.google.com/recaptcha/docs/v3#automatically_bind_the_challenge_to_a_button).
+    Warning: This is not a replacement for [official reCAPTCHA documentation](https://developers.google.com/recaptcha).
 
 # Feature set / ABI
 
--- a/plugins/samples/enable_recaptcha/plugin.rs
+++ b/plugins/samples/enable_recaptcha/plugin.rs
@@ -1,0 +1,208 @@
+// Copyright 2025 Google LLC
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+// [START serviceextensions_plugin_enable_recaptcha]
+
+// Warning: This plugin simply shows that by adding scripts and modifying a
+// button on a page, one could enable a reCAPTCHA challenge. This is not a
+// replacement for reading reCAPTCHA documentaion or following all user guide
+// instructions. Please follow official reCAPTCHA documentation at
+// https://developers.google.com/recaptcha. This plugin follows instructions for
+// "Automatically bind the challenge to a button" https://developers.google.com/recaptcha/docs/v3#automatically_bind_the_challenge_to_a_button.
+use lol_html::html_content::ContentType;
+use lol_html::*;
+use proxy_wasm::traits::*;
+use proxy_wasm::types::*;
+use std::cell::RefCell;
+use std::error::Error;
+use std::rc::Rc;
+
+proxy_wasm::main! {{
+    proxy_wasm::set_log_level(LogLevel::Trace);
+    proxy_wasm::set_http_context(|_, _| -> Box<dyn HttpContext> { Box::new(MyHttpContext::new())});
+}}
+
+struct MyOutputSink {
+    // Stores HTML as the rewriter parses and modifies HTML.
+    // Only stores sections of HTML that have not yet been seen back to client.
+    // After sending data to client, we clear output to avoid unnecessary memory
+    // growth.
+    // See comment in MyHttpContext about use of Rc<RefCell<T>>.
+    output_sink: Rc<RefCell<Vec<u8>>>,
+}
+
+impl OutputSink for MyOutputSink {
+    fn handle_chunk(&mut self, chunk: &[u8]) {
+        self.output_sink.borrow_mut().extend_from_slice(chunk);
+    }
+}
+
+struct MyHttpContext<'a> {
+    // Some member variables need to be wrapped in Rc<Refcell<T>> so that we can
+    // modify the variable in multiple places as well as share ownership.
+    // By using Rc, we are allowing multiple references to the same obeject to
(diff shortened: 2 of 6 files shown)


plugins/samples/enable_recaptcha/plugin.rs
```diff
@@ -0,0 +1,208 @@
...
+        MyHttpContext {
+            output: output.clone(),
+            completed_script_injection: completed_script_injection.clone(),
+            completed_button_modification: completed_button_modification.clone(),
+            rewriter: Some(HtmlRewriter::new(
+                Settings {
+                    element_content_handlers: vec![
+                        element!("head", move |el| {
+                            el.prepend(
+                            "\n<script src=\"https://www.google.com/recaptcha/api.js\"></script>\n\
```
jordan83, line 96: The format of the script we want here would look like this:

<script src="https://www.google.com/recaptcha/enterprise.js?render=KEY_ID"></script>

the key id should be accepted as plugin configuration, I think.

plugins/samples/enable_recaptcha/plugin.rs
```diff
@@ -0,0 +1,208 @@
...
+                            <script>function onLogin(token) {document.getElementById(\"login-form\").submit();}</script>",
```
jordan83, line 97: The only thing we actually want here would be the script injection. It's not a full solution as is, but we can build on it when the time comes.

plugins/samples/enable_recaptcha/plugin.rs
```diff
@@ -0,0 +1,221 @@
...
+
+impl Context for MyRootContext {}
+
+impl RootContext for MyRootContext {
+    fn on_configure(&mut self, _: usize) -> bool {
+        if let Some(config) = self.get_plugin_configuration() {
+            // Config file contains only the recaptcha key.
+            // Use of .unwrap() is fine here, since failure in on_configure
+            // should result in plugin crash.
+            let config_lines = String::from_utf8(config).unwrap();
```
jordan83, line 50: I don't know if this is already doing this or not but it would be good to trim whitespace.

plugins/samples/enable_recaptcha/tests.textpb
```diff
@@ -0,0 +1,8 @@
+test {
+  name: "Enable reCAPTCHA challenge"
+  num_chunks:10
```
jordan83, line 3: Can you explain a bit how the tests work? This tells the test runner to break the input body up into exactly 10 chunks? In a prod environment this is not deterministic right?

plugins/samples/enable_recaptcha/plugin.rs
```diff
@@ -0,0 +1,221 @@
...
+            .expect("Expected valid rewriter. Got None")
+            .end()?;
+        return Ok(());
+    }
+}
+
+impl<'a> Context for MyHttpContext<'a> {}
+
+impl<'a> HttpContext for MyHttpContext<'a> {
+    // If request header callback is invoked, body will not be compressed.
```
jordan83, line 161: What causes this to be invoked or not invoked? It doesn't appear to be configurable through the plugin configuration.

plugins/samples/enable_recaptcha/tests.textpb
dallen68: You're right. This breaks the body into 10 chunks. In prod this behavior is not deterministic. This just tests that the plugin can handle potentially ugly chunking boundaries.

plugins/samples/enable_recaptcha/plugin.rs
dallen68: Callbacks are invoked based on Envoy configuration. This repo does not replicate this functionality. This was only included in case users wanted the ability to disable compression.

plugins/samples/enable_recaptcha/plugin.rs
dallen68: I intentionally kept the handling of config data as bare-bones as possible. Users can trim whitespace if they find it necessary, but it isn't needed in this instance.

plugins/README.md
```diff
@@ -144,6 +144,9 @@ for your own plugin. Extend them to fit your particular use case.
 *   [Rewrite domains in html response body](samples/html_domain_rewrite/): Parse
     html in response body chunks and replace insances of "foo.com" with
     "bar.com" in `<a href=***>`.
+*   [Enable reCAPTCHA challenge on response body](samples/enable_recaptcha/):
```
jordan83, line 147: Please add docs for the compression configuration.

plugins/samples/enable_recaptcha/plugin.rs
```diff
@@ -0,0 +1,225 @@
...
+        return Ok(());
+    }
+}
+
+impl<'a> Context for MyHttpContext<'a> {}
+
+impl<'a> HttpContext for MyHttpContext<'a> {
+    // Use header 'stop-compression' to toggle body compression.
+    fn on_http_request_headers(&mut self, _: usize, _: bool) -> Action {
+        let stop_compression = self.get_http_request_header("Stop-compresssion");
```
jordan83, line 163: Strongly prefer this is in the plugin configuration rather than forcing users to modify web requests from their client to include this header. Fine for now I guess.

jordan83 (approved):

plugins/samples/enable_recaptcha/plugin.rs
dallen68: Ack

martijneken: I agree with @jordan83. This doesn't need to be a header-driven conditional. In the Service Extension resources configuration, a user may choose to invoke this request headers handler or not, so I propose we just remove the use of "Stop-compresssion" here, and replace it with a comment saying "invoke this handler to disable compression".

plugins/samples/enable_recaptcha/tests.textpb
```diff
@@ -0,0 +1,8 @@
...
+  response_body {
+    input { file: "response_body.data" }
+    result { body { file: "expected_response_body.data" } }
+  }
+}
```
martijneken, line 8: Please add a separate test for the accept-encoding behavior.

plugins/samples/enable_recaptcha/plugin.rs
martijneken: Decided to ditch the disable-compression feature, please remove.

plugins/samples/enable_recaptcha/tests.textpb
martijneken: Never mind, removing the feature.

plugins/samples/enable_recaptcha/plugin.rs
martijneken: Removing, per discussion

plugins/samples/enable_recaptcha/plugin.rs
```diff
@@ -0,0 +1,215 @@
...
+    // via content_handlers. The rewriter only writes completed sections of Html to
+    // the output sink(e.g., when the rewriter parses "<di", nothing will be
+    // sent to the output sink. If the next chunk was "v> <h1>foo</h", the
+    // rewriter would then write "<div><h1>foo" to the output sink) unless
+    // HtmlRewriter::end() is called. HtmlRewriter::end() results in all
+    // uncompleted Html being buffered in rewriter to be sent to output sink as
+    // if it were plain text.
+    rewriter: Option<HtmlRewriter<'a, MyOutputSink>>,
+    // True when plugin has added script to <head>.
+    completed_script_injection: Rc<RefCell<bool>>,
```
martijneken, line 105: Why use Rc<Refcell> here? Can this just be a plain bool?

martijneken (approved):
LGTM but please close open threads

PR #206, opened 2025-03-17
feat: Add a Go redirect plugin to include a bulk list of domains

pweiber:
This PR contains only Go version for the requirement

- Include a bulk list of domains
- Reference the bulk list from the plugin configuration data

### DIFF
--- a/plugins/samples/redirect_bulk/plugin.go
+++ b/plugins/samples/redirect_bulk/plugin.go
@@ -1,0 +1,139 @@
+// Copyright 2025 Google LLC
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+// [START serviceextensions_plugin_redirect_bulk]
+package main
+
+import (
+	"strings"
+
+	"github.com/proxy-wasm/proxy-wasm-go-sdk/proxywasm"
+	"github.com/proxy-wasm/proxy-wasm-go-sdk/proxywasm/types"
+)
+
+func main() {}
+
+func init() {
+	proxywasm.SetVMContext(&vmContext{})
+}
+
+type vmContext struct {
+	types.DefaultVMContext
+}
+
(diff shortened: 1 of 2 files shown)


plugins/samples/redirect_bulk/plugin.go
```diff
@@ -0,0 +1,139 @@
...
+		proxywasm.LogErrorf("Failed to get plugin configuration: %v", err)
+		return types.OnPluginStartStatusFailed
+	}
+
+	// Parse configuration
+	configStr := string(config)
+	mappings := make(map[string]string)
+
+	// Parse each line as "source_domain target_domain"
+	for _, line := range strings.Split(configStr, "\n") {
```
leonm1, line 70: This string conversion and the strings.Split will create a large number of strings that have to be garbage collected later.

Can you use the `bytes.Split(config, []byte("\n"))`, `bytes.TrimSpace()`, and `bytes.Fields`, the convert to string only on line 79 when inserting into the map? That will avoid creating any intermediate strings, and only create views into the single `config` byte slice.

plugins/samples/redirect_bulk/plugin.go
```diff
@@ -0,0 +1,139 @@
...
+		// Send a 301 response with the new location
+		headers := [][2]string{
+			{"Location", newURL},
+		}
+		body := []byte("Redirecting to " + newURL)
+		if err := proxywasm.SendHttpResponse(301, headers, body, -1); err != nil {
+			proxywasm.LogErrorf("Failed to send redirect response: %v", err)
+		}
+
+		return types.ActionPause // End stream and stop further processing
```
leonm1, line 133: It doesn't matter much what we return after `SendHttpResponse`, however, for service-extensions, ActionPause is currently unsupported. To avoid misleading users, can you change this to ActionContinue?

leonm1 (changes requested):

plugins/samples/redirect_bulk/plugin.go
pweiber: Done

plugins/samples/redirect_bulk/plugin.go
```diff
@@ -0,0 +1,139 @@
...
+		// Send a 301 response with the new location
+		headers := [][2]string{
+			{"Location", newURL},
+		}
+		body := []byte("Redirecting to " + newURL)
+		if err := proxywasm.SendHttpResponse(301, headers, body, -1); err != nil {
+			proxywasm.LogErrorf("Failed to send redirect response: %v", err)
+		}
+
+		return types.ActionContinue
```
mpwarres: Since the return value is now the same regardless of whether the if-clause is executed or not, suggest removing this `return` statement and just falling through to the one at the end of the function.

leonm1 (approved):

PR #217, opened 2025-04-09
Support local response from request body callback

leonm1:
The response body callback is not supported since at that point, the body may already be streaming to the downstream client. It is fair to send an immediate response from the request body callback; however.

### DIFF
--- a/plugins/test/framework.cc
+++ b/plugins/test/framework.cc
@@ -159,10 +159,12 @@ proxy_wasm::WasmResult TestHttpContext::sendLocalResponse(
     uint32_t response_code, std::string_view body_text,
     proxy_wasm::Pairs additional_headers, uint32_t grpc_status,
     std::string_view details) {
-  if (phase_ != proxy_wasm::WasmHeaderMapType::RequestHeaders &&
-      phase_ != proxy_wasm::WasmHeaderMapType::ResponseHeaders) {
+  if (current_callback_ != RequestHeaders &&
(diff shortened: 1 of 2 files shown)


plugins/test/framework.cc
```diff
@@ -193,7 +195,9 @@ TestHttpContext::Result TestHttpContext::SendRequestBody(std::string body) {
   current_callback_ = TestHttpContext::CallbackType::RequestBody;
   result_.body_status =
       onRequestBody(body_buffer_.size(), /*end_of_stream=*/false);
-  result_.body = body_buffer_.release();
+  if (!sent_local_response_) {
```
mpwarres, line 198: I wonder a little whether this should return BadArgument if sent_local_response_ is already true. Though that check is not currently performed in TuringStreamContext either.

mpwarres (approved):

martijneken:
> The response body callback is not supported

I actually think it should be supported: https://github.com/envoyproxy/envoy/blob/31d7b249a5a037c7300428faf92b30e44af89612/api/envoy/service/ext_proc/v3/external_processor.proto#L195-L201

I wrote an integration test where the client sees the immediate response (although not guaranteed):
https://github.com/envoyproxy/envoy/blob/31d7b249a5a037c7300428faf92b30e44af89612/test/extensions/filters/http/ext_proc/ext_proc_integration_test.cc#L2541

PR #224, opened 2025-04-24
feat: create a separate response for add header plugin

victorwoli:
Duplicate and create a separate recipe for adding response header from add header request


plugins/samples/add_request_header/plugin.cc
```diff
@@ -0,0 +1,32 @@
...
+// [START serviceextensions_plugin_add_header]
+#include "proxy_wasm_intrinsics.h"
+
+class MyHttpContext : public Context {
+ public:
+  explicit MyHttpContext(uint32_t id, RootContext* root) : Context(id, root) {}
+
+  FilterHeadersStatus onRequestHeaders(uint32_t headers,
+                                       bool end_of_stream) override {
+    // Always be a friendly proxy.
```
mpwarres, line 24: :-)

mpwarres (approved):

PR #231, opened 2025-05-13
feat: adding new go plugin sample - add device type

mateustd-ciandt:
This PR introduces a new `plugin.go` that implements a device type detection plugin based on the user agent header. 

It includes tests to ensure functionality. Future pull requests will include versions of this plugin in `.cc` and `.rs` for broader compatibility.

Performance Note: While the current string matching (`containsAny`) works well for most cases, we could potentially optimize further for high-throughput scenarios using more advanced algorithms like Aho-Corasick, Boyer-Moore, or hash search, but  I believe it would bring an unecessary complexy to the plugin.

PR #232, opened 2025-05-16
feat: adding new cpp plugin sample - add device type

mateustd-ciandt:
This PR introduces a new plugin.cc that implements a device type detection plugin based on the user agent header.

It includes tests to ensure functionality. We already have a [pull request for this plugin in Go](https://github.com/GoogleCloudPlatform/service-extensions/pull/231) and a future pull request will include the version of this plugin in `.rs` for broader compatibility.


plugins/samples/add_device_type/plugin.cc
```diff
@@ -0,0 +1,94 @@
...
+  std::string DetectDeviceType(const std::string& ua) const {
+    if (IsBot(ua)) return "bot";
+    if (IsTablet(ua)) return "tablet";
+    if (IsMobile(ua)) return "phone";
+    if (IsDesktop(ua)) return "desktop";
+    return "other";
+  }
+
+  bool IsBot(const std::string& ua) const {
+    const std::vector<std::string> bots = {
```
mpwarres, line 49: To avoid repeatedly constructing the same vectors for each incoming request, I'd suggest moving DetectDeviceType() and the various Is*() methods to a MyRootContext class (which is only instantiated once), and defining vector fields of that class to hold the various lists.

Or, if you want to hold the lists in static variables or function-local static variables, see https://google.github.io/styleguide/cppguide.html#Static_and_Global_Variables for safe idioms.

plugins/samples/add_device_type/plugin.cc
```diff
@@ -0,0 +1,94 @@
...
+    for (const auto& sub : subs) {
+      if (absl::StrContains(ua, sub)) return true;
+    }
+    return false;
+  }
+};
+
+static RegisterContextFactory register_StaticContext(
+    CONTEXT_FACTORY(MyHttpContext), ROOT_FACTORY(RootContext));
+// [END serviceextensions_plugin_device_type]
```
mpwarres, line 94: nit: terminate last line with a newline char

plugins/samples/add_device_type/tests.textpb
```diff
@@ -0,0 +1,92 @@
...
+  name: "BenchmarkDetection"
+  benchmark: true
+  request_headers {
+    input { 
+      header { key: "user-agent" value: "Mozilla/5.0 (Windows NT 10.0; Win64; x64) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/91.0.4472.124 Safari/537.36" } 
+      header { key: "host" value: "example.com" }
+    }
+    result { has_header { key: "x-device-type" value: "desktop" } }
+  }
+}
```
mpwarres, line 92: nit: terminate last line with a newline char

mpwarres (changes requested):
LGTM mod comments

plugins/samples/add_device_type/plugin.cc
mateustd-ciandt: sounds good, I gave it a try! thanks!

PR #237, opened 2025-05-29
feat: adding new go plugin sample - hmac token validation

mateustd-ciandt:
This Pull Request adds a GO WASM plugin that implements validating an MD5 signature with HMAC.

- Verifies token freshness against configurable expiry period.
- Computes HMAC-MD5 signatures of request data.
- Rejects requests with invalid/missing tokens.

It also includes BUILD files and test cases covering invalid formats, missing headers, signature validation, and token expiration.


plugins/samples/hmac_token_validation/BUILD
```diff
@@ -0,0 +1,16 @@
...
+    srcs = ["plugin.go"],
+)
+
+proxy_wasm_tests(
+    name = "tests",
+    plugins = [
+        ":plugin_go.wasm",
+    ],
+    tests = ":tests.textpb",
+)
```
mpwarres, line 16: nit: make sure that all files have a terminating newline char

mpwarres (commented):
Can this plugin sample be moved to plugins/samples/hmac_authtoken so it is alongside the C++ sample and can share the tests.textpb file with it?

PR #238, opened 2025-05-30
feat: create a Hello World samples for C++, GO and RUST plugins

victorwoli:
Create a "hello world" plugin - which responds to all requests with a 200 and a hello world in the body


plugins/samples/hello_world/plugin.go
```diff
@@ -0,0 +1,53 @@
...
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+// [START serviceextensions_plugin_hello_world]
+package main
+import (
+	"github.com/proxy-wasm/proxy-wasm-go-sdk/proxywasm"
+	"github.com/proxy-wasm/proxy-wasm-go-sdk/proxywasm/types"
+)
+func main() {}
```
mpwarres, line 20: nit: add a blank line between each type/func definition

victorwoli:
@mpwarres I believe all the suggested fixes are done, please let me know if there is anything else.

PR #244, opened 2025-06-22
Add support for running background streams in plugin benchmarks

### LINKED ISSUE #191: Plugins tester: improve memory stats
The memory stat (WasmMemoryB) could use further improvement since it

1. Doesn't account for stream parallelism (any VM/thread must serve 1000 concurrent streams):

    In DynamicTest::BenchStreamLifecycle, we should hold a flag-configurable number of open streams (i.e. 1000).

    In DynamicTest::BenchHttpHandlers, we should additionally interleave the streams' callbacks "randomly" onto one wasm thread, so as to measure a sum-of-peaks set of allocations. This would definitely screw up wall timing; could we still measure CPU time per stream or would this need to be a separate memtest?

    See also: https://github.com/GoogleCloudPlatform/service-extensions/blob/df4615ecce7681ebbea45e856cb754c45ad13510/plugins/test/dynamic_test.cc#L415-L424

2. Doesn't give insight into wasm-side allocations. Consider special-casing some calls into one or more wasm runtimes to grab wasm-side heap stats. For v8, `V8::Isolate::GetHeapStatistics` would return the actual memory malloc'd by the plugin.

mpwarres:
Adds two command-line flags to the plugin tester:

--num_background_streams
    Specifies the number of background streams to instantiate in benchmarks.

--background_stream_advance_rate
    If background streams are configured, specifies how many background
    streams to advance the state of in each benchmark cycle.

Background streams are instantiated and advanced outside of the benchmark timer. They pass the same headers and body inputs to the plugin that are specified for the test case being benchmarked.

The num_background_streams and background_stream_advance_rate values can also be configured by setting the corresponding fields of the service_extensions.samples.pb.Env protobuf.

Fixes #191


mpwarres:
This PR is obsoleted by #248.

PR #248, opened 2025-07-07
feat: add support for running background streams in plugin benchmarks

### LINKED ISSUE #191: Plugins tester: improve memory stats

mpwarres:
Adds two command-line flags to the plugin tester:

--num_additional_streams
    Specifies the number of additional streams to instantiate in benchmarks.

--additional_stream_advance_rate
    If additional streams are configured, specifies how many background
    streams to advance the state of in each benchmark cycle.

Additional streams are instantiated and advanced outside of the benchmark timer. They pass the same headers and body inputs to the plugin that are specified for the test case being benchmarked.

Example usage:
```
$ cd plugins/
$ bazelisk build ...
$ bazel-bin/test/runner \
    --proto=samples/add_response_header/tests.textpb \
    --plugin=bazel-bin/samples/add_response_header/plugin_cpp.wasm \
    --num_additional_streams=1000 \
    --additional_stream_advance_rate=5
```

The num_additional_streams and additional_stream_advance_rate values can also be configured by setting the corresponding fields of the service_extensions.samples.pb.Env protobuf.

Fixes #191

### DIFF
--- a/plugins/test/dynamic_test.cc
+++ b/plugins/test/dynamic_test.cc
@@ -18,6 +18,7 @@
 
 #include <boost/filesystem/path.hpp>
 
+#include "absl/random/distributions.h"
 #include "absl/strings/str_cat.h"
 #include "absl/strings/substitute.h"
 #include "benchmark/benchmark.h"
@@ -34,6 +35,7 @@
 
 namespace service_extensions_samples {
 namespace {
+
 // Helper to read data from a path which may be relative to the test config.
 absl::StatusOr<std::string> ReadContent(const std::string& path,
                                         const pb::Env& env) {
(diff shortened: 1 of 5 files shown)


plugins/test/dynamic_test.cc
```diff
@@ -394,6 +550,16 @@ void DynamicTest::BenchHttpHandlers(benchmark::State& state) {
     std::vector<std::string> request_body_chunks_copies = *request_body_chunks;
     std::vector<std::string> response_body_chunks_copies =
         *response_body_chunks;
+
+    // Advance a fraction of background streams, if present.
```
leonm1, line 554: This is happening while timing is paused. Is the intention such that background streams will load the VM, but not impact the CPU or wall time measurements of the test cases?

I suppose that's necessary, since otherwise the benchmark's values wouldn't always converge and wouldn't be a strict measurement of the test case.

Could you add a comment to the new proto fields to explain the impact they have on measurements? I.e. something like "The cpu time of the background streams is not measured in the microbenchmark. Rather, the background streams allow us to test the cost of a single stream while the plugin is loaded with other streams, which may catch memory usage issues or performance issues with algorithmic inefficiencies with large number of in-flight requests"

plugins/test/dynamic_test.cc
```diff
@@ -118,6 +120,132 @@ class LogTestBounds {
...
+};
+
+absl::Status BackgroundStream::Advance() {
+  while (true) {  // cycle through actions until one is performed
+    switch (next_action_) {
+      case NextAction::kCreate:
+        stream_ = std::make_unique<TestHttpContext>(handle_);
+        if (request_body_chunks_ != nullptr) {
+          remaining_request_body_chunks_.assign(request_body_chunks_->begin(),
+                                                request_body_chunks_->end());
```
martijneken, line 174: I don't remember the ownership model of body chunks, and I'm a little worried that this class refers to chunks that may be later consumed by the "main streams". Is that a real issue?

Either way please add a comment somewhere in this class explaining the ownership model.

plugins/test/dynamic_test.cc
```diff
@@ -384,6 +521,25 @@ void DynamicTest::BenchHttpHandlers(benchmark::State& state) {
...
+  if (env_.num_background_streams() > 0) {
+    for (int i = 0; i < env_.num_background_streams(); ++i) {
+      background_streams.emplace_back(
+          handle, request_headers.has_value() ? &*request_headers : nullptr,
+          response_headers.has_value() ? &*response_headers : nullptr,
+          &*request_body_chunks, &*response_body_chunks);
+      // Advance twice, once to create the TestHttpContext and once to perform
+      // an HTTP handler callback.
+      BM_RETURN_IF_ERROR(background_streams.back().Advance());
+      BM_RETURN_IF_ERROR(background_streams.back().Advance());
```
martijneken, line 534: Is the second Advance() really needed here? I assume the first HTTP handler could also be invoked below in the main loop.

plugins/test/runner.proto
```diff
@@ -147,7 +147,13 @@ message Env {
   // Log file to receive plugin output. Can be /dev/std{out,err}.
   string log_path = 5;
   // Fixed timestamp in Unix seconds. If unset it will be the Unix epoch.
-  uint64 time_secs = 6;
+  optional uint64 time_secs = 6;
+  // Number of background streams to run in benchmarks. If unset, no background
+  // streams will be run.
+  optional uint64 num_background_streams = 7;
```
martijneken, line 153: naming bikeshed: The term "background" implies multithreading to me, and happily that is not how this works. Maybe something like 'other_streams' or 'memory_pressure_streams' or 'unmeasured_streams'?

plugins/test/runner.proto
```diff
@@ -147,7 +147,13 @@ message Env {
...
+  // If there are background streams, this controls the number of background
+  // streams that are advanced per iteration of the benchmark loop.
+  optional uint64 background_stream_advance_rate = 8;
```
martijneken, line 156: The concept of 'advance rate' is a bit unintuitive to me. I think it means 'for every 1 CPU-measured cycle, we also run N CPU-unmeasured cycle-phases'. What's tricky is that this doesn't control the number of cycles in total, nor the number of phases executed in the main cycle.

I don't have any great suggestions here other than expanding the comment and/or calling it a 'ratio'.

plugins/test/runner_main.cc
```diff
@@ -44,6 +46,10 @@ ABSL_FLAG(service_extensions_samples::pb::Env::LogLevel, loglevel,
           "Override log_level.");
 ABSL_FLAG(bool, test, true, "Option to disable config-requested tests.");
 ABSL_FLAG(bool, bench, true, "Option to disable config-requested benchmarks.");
+ABSL_FLAG(uint64_t, num_background_streams, 0,
+          "Number of background streams to run in benchmarks.");
+ABSL_FLAG(std::optional<uint64_t>, background_stream_advance_rate, std::nullopt,
```
martijneken, line 51: Why is this one optional, but the one above isn't? This must be >0 to make sense, so ignoring a default value of 0 seems OK to me. (no need to support disabling the advance feature IMO, since we have the StreamLifecycle benchmark to measure that behavior)

plugins/test/dynamic_test.cc
mpwarres: > I don't remember the ownership model of body chunks, and I'm a little worried that this class refers to chunks that may be later consumed by the "main streams". Is that a real issue?

It shouldn't be an issue--DynamicTest::BenchHttpHandlers() creates copies of body data with each loop iteration ([code](https://github.com/GoogleCloudPlatform/service-extensions/blob/09182c7881b41ae4769be36547e93b52531949ee/plugins/test/dynamic_test.cc#L394-L395)), while BackgroundStream holds a pointer to the original body data vector that is only used as the source for copies.

Added a comment to the fields about ownership.

plugins/test/dynamic_test.cc
```diff
@@ -384,6 +521,25 @@ void DynamicTest::BenchHttpHandlers(benchmark::State& state) {
       PrepBodyCallbackBenchmark(cfg_, cfg_.response_body());
   BM_RETURN_IF_ERROR(response_body_chunks.status());
 
+  std::vector<BackgroundStream> background_streams;
+  if (env_.num_background_streams() > 0) {
```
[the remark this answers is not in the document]
mpwarres: Removed

plugins/test/dynamic_test.cc
mpwarres: While it's probably the case that the benchmark loop would execute enough times to advance all streams to the point where they actually perform an HTTP callback and incur whatever memory usage comes with it, this ensures that, removing some (slight) unpredictability. While I agree it's a little clunky aesthetically, I don't see any other downside, so would prefer to let function take precedence over form here.

plugins/test/dynamic_test.cc
mpwarres: > This is happening while timing is paused. Is the intention such that background streams will load the VM, but not impact the CPU or wall time measurements of the test cases?

Yes, exactly.

> Could you add a comment to the new proto fields to explain the impact they have on measurements?

Sure--added the text you suggested. Thanks!

plugins/test/runner.proto
mpwarres: Changed to "additional streams"

plugins/test/runner.proto
mpwarres: I started to rename it to "ratio", but then that field a little weird to me since the value is an int, and ratio seems like it should allow floats / fractions. So I left the name as-is but expanded the comment.

plugins/test/runner_main.cc
mpwarres: Done

mpwarres (commented):
Thanks for the reviews!

PR #252, opened 2025-07-23
feat: improving wasm tracing for string params

### LINKED ISSUE #195: Plugins tester: improve wasm tracing for string params
The --min_log_level=TRACE param shows the ABI boundary calls (see below). Often these contain pointers to strings in wasm linear memory. When these are strings (addresses+lengths), consider printing the string content directly, for improved legibility.

```
TRACE from integration: [host<-vm] proxy_on_configure return: 1
TRACE from integration: [host->vm] proxy_on_context_create(2, 1)
TRACE from integration: [vm->host] env.proxy_log(2, 5247632, 65)
TRACE from integration: [vm<-host] env.proxy_log return: 0
TRACE from integration: [host<-vm] proxy_on_context_create return: void
TRACE from integration: [host->vm] proxy_on_request_headers(2, 0, 0)
TRACE from integration: [vm->host] env.proxy_log(2, 5247632, 81)
TRACE from integration: [vm<-host] env.proxy_log return: 0
TRACE from integration: [host<-vm] proxy_on_request_headers return: 0
TRACE from integration: [host->vm] proxy_on_response_headers(2, 0, 0)
TRACE from integration: [vm->host] env.proxy_log(2, 5247632, 83)
TRACE from integration: [vm<-host] env.proxy_log return: 0
TRACE from integration: [host<-vm] proxy_on_response_headers return: 0
TRACE from integration: [host->vm] proxy_on_done(2)
TRACE from integration: [vm->host] env.proxy_log(2, 5247560, 61)
```

mateustd-ciandt:
Fixes #195

This PR updates the logging mechanism in TestContext::log by enhancing the output format. The previous implementation simply printed the log message, while the new version provides more context by displaying the log level, memory address of the message, and its size.

This change will improve our debugging and tracing capabilities, making it easier to monitor logs and evaluate performance during development. Feel free to modify it as necessary!

### DIFF
--- a/plugins/test/framework.cc
+++ b/plugins/test/framework.cc
@@ -62,10 +62,14 @@ uint64_t TestContext::getMonotonicTimeNanoseconds() {
   return absl::ToUnixNanos(options().clock_time);
 }
 proxy_wasm::WasmResult TestContext::log(uint32_t log_level,
-                                        std::string_view message) {
+  std::string_view message) {
   logging_bytes_ += message.size();
   if (wasmVm()->cmpLogLevel(proxy_wasm::LogLevel::trace)) {
-    std::cout << "TRACE from testcontext: [log] " << message << std::endl;
+    std::cout << "TRACE from integration: [vm->host] env.proxy_log(" 
+    << log_level << ", " 
+    << reinterpret_cast<uint64_t>(message.data()) << ", " 
+    << message.size() << ") \"" 
+    << message << "\"" << std::endl;
   }
   if (wasmVm()->cmpLogLevel(static_cast<proxy_wasm::LogLevel>(log_level))) {
     phase_logs_.emplace_back(message);


plugins/test/framework.cc
```diff
@@ -62,10 +62,14 @@ uint64_t TestContext::getMonotonicTimeNanoseconds() {
...
 }
 proxy_wasm::WasmResult TestContext::log(uint32_t log_level,
-                                        std::string_view message) {
+  std::string_view message) {
   logging_bytes_ += message.size();
   if (wasmVm()->cmpLogLevel(proxy_wasm::LogLevel::trace)) {
-    std::cout << "TRACE from testcontext: [log] " << message << std::endl;
+    std::cout << "TRACE from integration: [vm->host] env.proxy_log(" 
+    << log_level << ", " 
+    << reinterpret_cast<uint64_t>(message.data()) << ", "
```
mpwarres, line 70: I'm not sure how useful it is to display the address and size of the message string, given that the value of the string is already output. Did you have a case in mind where this information would be valuable?

martijneken: Friendly ping on this PR. I had a use for it today (testing/inspecting a plugin I did not write and trying to understand its behaviors).

PR #258, opened 2025-08-06
feat: Add a cpp plugin to include a custom error page with the traceid

mateustd-ciandt:
Adding a plugin in c++ that surfaces traceID in custom error pages.
We're checking for both Google Cloud trace header and W3C header.
Also adding a BUILD file and a tests.textpb file.


plugins/samples/error_page_with_traceid/BUILD
```diff
@@ -0,0 +1,23 @@
...
+    ],
+)
+
+proxy_wasm_tests(
+    name = "tests",
+    plugins = [
+        ":plugin_cpp.wasm",
+    ],
+    tests = ":tests.textpb",
+)
```
mpwarres, line 23: nit: add newline at end of file

plugins/samples/error_page_with_traceid/plugin.cc
```diff
@@ -0,0 +1,146 @@
...
+    return FilterHeadersStatus::Continue;
+  }
+
+  FilterHeadersStatus onResponseHeaders(uint32_t headers,
+                                        bool end_of_stream) override {
+    const auto status_header = getResponseHeader(":status");
+    if (!status_header) return FilterHeadersStatus::Continue;
+
+    status_code_ = status_header->toString();
+    const int status_code = std::stoi(status_code_);
```
mpwarres, line 70: Suggest using absl::SimpleAtoi. IIRC std::stoi can throw an exception, which would fail the plugin.

plugins/samples/error_page_with_traceid/plugin.cc
```diff
@@ -0,0 +1,146 @@
...
+      }
+      return trace_context;
+    }
+
+    // Try W3C Trace Context standard
+    const auto w3c_trace = getRequestHeader("traceparent");
+    if (w3c_trace && !w3c_trace->view().empty()) {
+      const std::string trace_context = w3c_trace->toString();
+      
+      // Format: version-trace_id-parent_id-flags
```
mpwarres, line 113: Suggest using regex to match, with capture group to extract the trace ID part. The regex will also allow this to verify that proper hexdigit chars are used.

plugins/samples/error_page_with_traceid/plugin.cc
```diff
@@ -0,0 +1,146 @@
...
+    return "not-available";
+  }
+
+  std::string status_code_;
+  std::string trace_id_{"not-available"};  // Default value
+};
+
+static RegisterContextFactory register_StaticContext(
+    CONTEXT_FACTORY(MyHttpContext), ROOT_FACTORY(RootContext));
+// [END serviceextensions_plugin_error_page_with_traceid]
```
mpwarres, line 146: nit: add newline char at end of file

plugins/samples/error_page_with_traceid/tests.textpb
```diff
@@ -0,0 +1,95 @@
...
+  response_headers {
+    input {
+      header { key: ":status" value: "500" }
+    }
+    result {
+      immediate { http_status: 500 }
+      body { regex: "(?s).*trace-without-span-id.*" }
+    }
+  }
+}
```
mpwarres, line 95: nit: add newline char at end of file

mpwarres (commented):
LGTM mod comments

plugins/samples/error_page_with_traceid/plugin.cc
mateustd-ciandt: Done! Added all the newlines, also using hex and using absl::SimpleAtoi. Let me know if we could change something else! Thanks!

mpwarres: Thanks! A few follow-on comments:

1. Use the RE2 regex library instead of std::regex--you can model after the existing use of RE2 in the check_pii plugin [here](https://github.com/GoogleCloudPlatform/service-extensions/blob/03ef4ea645b84086420958b23893942f620401c4/plugins/samples/check_pii/plugin.cc#L17).
2. Similar to the example in (1), initialize the regex during plugin startup (i.e. in onStart() or onConfigure()), since regex initialization is expensive, so you don't want to do it per-request.
3. Use the regex to extract the desired part of the string using a capture group, e.g.:

   ```
   // in onStart() or onConfigure()
   w3c_trace_context_regex_.emplace(R"00-([0-9a-fA-F]{32})-[0-9a-fA-F]{16}-[0-9a-fA-F]{2}");

   // elsewhere
   absl::Status<std::string> extractTraceId(std::string_view trace_context) {
     if (std::string trace_id; RE2::FullMatch(w3c_trace.view(), w3c_trace_context_regex_, &trace_id)) {
       return trace_id;
     } else {
       return absl::InvalidArgumentError("invalid W3C trace context");
     }
   }
   ```

   Also, as illustrated in the snippet above, I'd recommend having extractTraceId() return an absl::Status<std::string> rather than using a magic string value to indicate failure.

mpwarres (approved):
LG, thanks!

PR #260, opened 2025-08-13
feat: add a basic l4 ext_proc example python grpc server

botengyao:
Added a basic l4 ext_proc loop_back python gRPC server.

This is using the network (TCP) filter [Ext_proc service](https://github.com/envoyproxy/envoy/blob/85e2494b39b4a41183c2c6bf99da1329ee1f7d5a/api/envoy/service/network_ext_proc/v3/network_external_processor.proto) from Envoy.

This is different from the http ext_proc interface.


callouts/python/README.md
```diff
@@ -4,6 +4,7 @@ Files using Copyright 2023 Google LLC & Apache License Version 2.0:
 
 * [service_callout.py](./extproc/service/callout_server.py)
```
mpwarres, line 5: Should ./extproc/service/network_callout_server.py also be included in this list? If you add it here, then it's probably worth a one-line description next to each file, where you could state that callout_server.py is for HTTP-layer callouts, while network_callout_server is for network-layer callouts.

callouts/python/README.md
```diff
@@ -4,6 +4,7 @@ Files using Copyright 2023 Google LLC & Apache License Version 2.0:
...
 * [callout_tools.py](./extproc/service/callout_tools.py)
+* [command_line_tools.py](./extproc/service/command_line_tools.py)
 
 # Requirements
```
mpwarres, line 10: In the "Quick start" section, comment "The minimal operation of this Python-based ext_proc server" needs to be updated (i.e. at least made plural) since this directory now houses two ext_proc server implementations.

callouts/python/README.md
```diff
@@ -360,3 +361,31 @@ docker build \
   -f ./extproc/example/jwt_auth/Dockerfile \
   -t service-callout-jwt-example-python .
 ```
+
+# Docker with TCP extension
```
mpwarres, line 365: Not suggested it has to be added in this PR, but a bunch of the documentation up to this point in the file (e.g. the "Developing Callouts" section) is specific to the HTTP-level callout server, and doesn't address L4 callouts. How about tweaking the title of that section to make clear that it is specific to L7 callouts, and adding a placeholder for a similar section for L4 callouts?

callouts/python/extproc/l4_example/Dockerfile
```diff
@@ -0,0 +1,64 @@
+# Copyright 2024 Google LLC.
```
mpwarres, line 1: nit: 2025

callouts/python/extproc/l4_example/Dockerfile
```diff
@@ -0,0 +1,64 @@
...
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+
+# Build the proto libraries from the envoy git repo using buf.
+FROM bufbuild/buf as BUF
+RUN apk add git
+COPY ./buf.gen.yaml ./buf.gen.yaml
+RUN buf -v generate https://github.com/envoyproxy/envoy.git#subdir=api --path envoy/service/network_ext_proc/v3/network_external_processor.proto --include-imports
```
mpwarres, line 20: This Dockerfile looks almost identical to callouts/python/extproc/example/Dockerfile, with the exception of (maybe only?) this line.

Not saying it needs to be done in this PR, but I wonder if the same Dockerfile could be used for the L4 and L7 examples, parameterized using a Docker --build-arg argument.

callouts/python/extproc/l4_example/network_basic/network_service_callout_example.py
```diff
@@ -0,0 +1,79 @@
+# Copyright 2024 Google LLC.
```
mpwarres, line 1: nit: 2025

callouts/python/extproc/service/network_callout_server.py
```diff
@@ -0,0 +1,364 @@
...
+    )
+
+    return response
+
+  def on_read_data(
+      self,
+      data: bytes,
+      end_of_stream: bool,
+      context: ServicerContext,
+  ) -> Tuple[bytes, bool]:
```
mpwarres, line 280: Consider using a namedtuple or dataclass with named fields `processed_data` and `modified` rather than a raw Tuple, to make the API more self-documenting and error-proof. See go/python-tips/019 for more info on these.

callouts/python/extproc/service/network_callout_server.py
```diff
@@ -0,0 +1,364 @@
...
+    """
+    # Default: pass through unchanged
+    return data, False
+
+  def on_write_data(
+      self,
+      data: bytes,
+      end_of_stream: bool,
+      context: ServicerContext,
+  ) -> Tuple[bytes, bool]:
```
mpwarres, line 301: Same comment as for on_read_data: consider using a namedtuple or dataclass.

callouts/python/extproc/service/network_callout_server.py
```diff
@@ -0,0 +1,364 @@
...
+    """Close the sockets of all servers, and trigger shutdowns."""
+    if self._health_check_server:
+      self._health_check_server.server_close()
+      self._health_check_server.shutdown()
+      logging.info('Health check server stopped.')
+
+    if self._callout_server:
+      self._callout_server.stop()
+
+  def _loop_server(self) -> None:
```
mpwarres, line 190: Not suggesting a change for this PR, but noticing that there is a fair amount of repeated boilerplate between this and the L7 callout_server.py, maybe a future PR could refactor to share more code.

callouts/python/extproc/l4_example/network_basic/network_service_callout_example.py
```diff
@@ -0,0 +1,79 @@
...
+
+  def on_read_data(
+      self,
+      data: bytes,
+      end_of_stream: bool,
+      context: ServicerContext,
+  ) -> Tuple[bytes, bool]:
+    """Process data from client to server (read path).
+    
+    Override this method to implement custom processing logic.
```
mpwarres, line 36: The current doc comment makes sense for the superclass, but here for this example, I think the doc comment for on_read_data should just describe what the example on_read_data does (i.e. just logs a message), and doesn't need to repeat the specification of Args and Returns, unless some linter requires it.

callouts/python/extproc/l4_example/network_basic/network_service_callout_example.py
```diff
@@ -0,0 +1,79 @@
...
+
+  def on_write_data(
+      self,
+      data: bytes,
+      end_of_stream: bool,
+      context: ServicerContext,
+  ) -> Tuple[bytes, bool]:
+    """Process data from server to client (write path).
+    
+    Override this method to implement custom processing logic.
```
mpwarres, line 58: Same comment as for on_read_data, just state what the example impl does, and no need to repeat the general interface doc for the method.

mpwarres (approved):
LGTM mod comments. Thanks!

callouts/python/extproc/service/network_callout_server.py
botengyao: done, dataclass makes sense.

callouts/python/extproc/service/network_callout_server.py
botengyao: done.

callouts/python/README.md
botengyao: agreed, added more L7 marker, and will add more for the L4 at the end of this Readme.

callouts/python/README.md
botengyao: done.

botengyao (commented):
thanks for the review!

callouts/python/extproc/l4_example/network_basic/network_service_callout_example.py
botengyao: good catch, done.

mpwarres (approved):
LG, thanks!

PR #273, opened 2025-09-10
chore: Add licenses for recaptcha plugin

dallen68:
chore: Add licenses for recaptcha plugin


martijneken (approved):
Wow, that's quite a license dump.

Is there a Rust tool that generates this? Please say how you generated it in the PR description so that future plugin authors may benefit from your process.

PR #288, opened 2025-11-10
feat: Add a GO plugin for GAM ad insertion into stream

mateustd-ciandt:
This plugin enables configurable Google Ad Manager ad injection into HTML streams.

It supports multiple ad positions with flexible placement options (before or after any HTML tag), follows standard GAM GPT implementation, and includes tests. The solution prevents third-party domain blocking by serving ads from the same parent domain as the content.

We also have a pull request for the cpp version [here](https://github.com/GoogleCloudPlatform/service-extensions/pull/285).


dallen68 (changes requested):
Can you update this PR with the reading from config and testing changes I requested in #285?

PR #293, opened 2025-11-24
feat: Adding new drafting token with JWT Rust plugin 

victorwoli:
Adding new drafting token with JWT plugin. This will be a Google Cloud service extension that can generate and validate JWT tokens.


plugins/samples/drafting_jwt_token/plugin.rs
```diff
@@ -0,0 +1,460 @@
...
+                401,
+                &VerifyResponse {
+                    valid: false,
+                    message: "Invalid Authorization format".to_string(),
+                }
+            );
+            return Action::Pause;
+        }
+
+        let token = &auth_header[7..];
```
martijneken: This strips the "Bearer " prefix? Worth a comment.

PR #295, opened 2025-11-28
feat: Add a go plugin to add geo query string to URL

mateustd-ciandt:
Adding a new GO plugin in order to modulate content based on the country value.
This value is added as a query string name value pair in the URL.
The query string value is also available in the normal GCP logs.

C++ version PR: https://github.com/GoogleCloudPlatform/service-extensions/pull/289
Rust version PR: https://github.com/GoogleCloudPlatform/service-extensions/pull/294


mpwarres (commented):
LGTM mod comment. Thanks!

PR #251, opened 2025-07-15
feat: adding load tests for the go callouts samples

mateustd-ciandt:
Adding load test scripts for all GO callout scenarios.

Also adding a README.md file for each new script. All scripts include proper error handling and cleanup. Documentation covers setup, execution and framework used.

PR #298, opened 2025-12-03
feat: Adding Load testing framework for callouts

pweiber:
Load testing framework aiming to replace the PRs:

#253 #241 #251

Also cover for the load testing samples inside

#185 - C++
#216 - Rust

The framework includes a single source of configuration [test-config.json] with multiple settings and the results can be exported as json, prometheus and influxdb formats for better analysis for the end user.

The framework runs in a controlled docker environment so all the tests run is isolated.


callouts/tests/load-testing/Dockerfile.load-tester
```diff
@@ -0,0 +1,39 @@
...
+    && rm /tmp/ghz.tar.gz /tmp/ghz.sha256
+
+WORKDIR /workspace
+
+COPY run-test.sh ./
+COPY lib/ ./lib/
+COPY config/test-config.json ./config/
+RUN chmod +x run-test.sh && chmod +x lib/*.sh
+
+COPY proto/ ./proto/
```
mpwarres: Is it possible to extract these files from Envoy and Google protobuf sources rather than keeping local copies that are unattributed and could drift out of sync, or do the local copies have modifications / simplifications?

mpwarres (approved):
LGTM mod one question. Thanks!

PR #306, opened 2026-01-16
feat: Add a cpp plugin for geo routing

mateustd-ciandt:
This PR adds a C++ plugin that reads the client's country code from load balancer properties (`source.client_region`) and sets it as an `x-country-code` request header. 
This header enables URL Map routing rules to direct traffic to geographically appropriate backend services. 

The plugin includes security measures to prevent header spoofing and is optimized for performance (~55-58μs latency per request following the [quotas and limits for Service Extensions.](https://docs.cloud.google.com/service-extensions/docs/quotas)).


mpwarres (approved):

plugins/samples/geo_directional_origin/plugin.cc
```diff
@@ -0,0 +1,56 @@
...
+// by URL Map routing rules to direct traffic to geographically appropriate
+// backend services.
+class MyHttpContext : public Context {
+ public:
+  explicit MyHttpContext(uint32_t id, RootContext* root) : Context(id, root) {}
+
+  FilterHeadersStatus onRequestHeaders(uint32_t headers,
+                                       bool end_of_stream) override {
+    // Retrieve the client's geographic region from load balancer properties.
+    auto country_property = getProperty({"request", "client_region"});
```
mpwarres: Similar to #348, can you update this to be "source" instead of "request"?

mateustd-ciandt: yep, will update these PRs as well!

mateustd-ciandt: Done! I also updated the tests.textpb file here to make sure we don't merge an old one by accident, should be good now!

PR #308, opened 2026-01-21
feat: Add Rust Plugins - Set/re-set cookies

victorwoli:
This plugin lets you set, reset, or delete a domain cookie before the request reaches the Cloud/Media CDN cache. You can create session cookies or time-based cookies using Apache log format. If a cookie already exists from the origin, the plugin can overwrite it or remove the set-cookie header entirely.


plugins/samples/set_reset_cookie/plugin.rs
```diff
@@ -0,0 +1,344 @@
...
+}
+
+impl MyHttpContext {
+    fn parse_request_cookies(&self) -> Vec<(String, String)> {
+        let mut cookies = Vec::new();
+        if let Some(cookie_header) = self.get_http_request_header("Cookie") {
+            if cookie_header.is_empty() {
+                return cookies;
+            }
+            for pair in cookie_header.split("; ") {
```
mpwarres, line 225: To be more tolerant of variable whitespace (which technically is allowed), split on ";" (no whitespace) and then trim whitespace from the resulting value, e.g.:

```
for pair in cookie_header.split(';') {
    let pair = pair.trim();
    if let Some(eq_pos) = pair.find('=') {
       // ...
    }
}
```

plugins/samples/set_reset_cookie/plugin.rs
```diff
@@ -0,0 +1,344 @@
...
+        cookies
+    }
+
+    fn process_cookie_deletions(&self, mut request_cookies: Vec<(String, String)>) {
+        let mut modified = false;
+        for config in self.cookie_configs.iter() {
+            if config.operation != OP_DELETE {
+                continue;
+            }
+            if let Some(pos) = request_cookies.iter().position(|(name, _)| name == &config.name) {
```
mpwarres, line 240: Can there be multiple cookies that match the name?

mpwarres (commented):
LGTM mod minor comments. Also, please add a README.md file similar to other plugin samples. Thanks!

PR #311, opened 2026-01-29
feat: Add a GO plugin for geo routing

mateustd-ciandt:
This PR adds a GO plugin that reads the client's country code from load balancer properties (source.client_region) and sets it as an x-country-code request header.
This header enables URL Map routing rules to direct traffic to geographically appropriate backend services.


mpwarres (approved):

mateustd-ciandt:
@mpwarres we should be good now!

plugins/samples/geo_directional_origin/plugin.go
```diff
@@ -0,0 +1,73 @@
...
+// [START serviceextensions_plugin_geo_routing]
+package main
+
+import (
+	"github.com/proxy-wasm/proxy-wasm-go-sdk/proxywasm"
+	"github.com/proxy-wasm/proxy-wasm-go-sdk/proxywasm/types"
+)
+
+const (
+	clientRegionProperty = "request.client_region"
```
mpwarres: Similar to #348, can you update this to be `source.client_region`?

mateustd-ciandt: Done! I also updated the tests.textpb file here to make sure we don't merge an old one by accident, should be good now!

mpwarres (approved):

PR #333, opened 2026-04-08
fix: Add default suppression for cert and private key paths

jstraceski:
We want to use the class level default if no path is specified:

```
import argparse
... 
... parser = argparse.ArgumentParser()
... 
... parser.add_argument('--foo', type=str)
... parser.add_argument('--bar', type=str, default=argparse.SUPPRESS)
... 
... 
... def test(foo='hello', bar='goodbye'):
...   print(foo, bar)
... 
... 
... test(**vars(parser.parse_args()))
... 
None goodbye
```
i.e. `goodbye` over `None`.
This could be accomplished here, within the class, or both as None is not a valid option either.

Fixes #<issue_number_goes_here>

- [ ] Tests pass


mpwarres (approved):
Thanks for fixing this!

PR #341, opened 2026-05-08
fix: Improve YAML parser

martijneken:
Fixes #307

Walks the YAML and Proto formats at the same time, uses proto reflection to understand field types. Fully generic.

- [x] Tests pass

### DIFF
--- a/plugins/test/yaml_proto_converter.cpp
+++ b/plugins/test/yaml_proto_converter.cpp
@@ -35,41 +35,20 @@ namespace {
 // Forward declaration
 google::protobuf::Value ConvertYamlNodeToProtoValue(const YAML::Node& node, const std::string& key = "");
 
-// Normalizes a header string "Key: Value" or map {"key": K, "value": V} or {K: V} into a Struct Value.
-google::protobuf::Value NormalizeHeader(const YAML::Node& node) {
+// Normalizes a scalar header string "Key: Value" or "Key" into a Struct Value.
+google::protobuf::Value NormalizeHeaderScalar(const std::string& header_str) {
   google::protobuf::Value header_val;
   auto* struct_fields = header_val.mutable_struct_value()->mutable_fields();
 
-  if (node.IsMap()) {
-    if (node["key"] || node["value"]) {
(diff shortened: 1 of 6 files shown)


plugins/test/yaml_proto_converter.cpp
```diff
@@ -16,328 +16,199 @@
...
-  for (const auto& yaml_invocation : yaml_invocations) {
-    Invocation* proto_invocation = proto_invocations->Add();
-    if (absl::Status status = ConvertInvocation(yaml_invocation, proto_invocation); !status.ok()) return status;
+    const YAML::Node& value_node = it.second;
+    if (field->is_repeated()) {
+      if (!value_node.IsSequence()) {
+        return absl::InvalidArgumentError(absl::StrCat("Expected a Sequence for repeated field '", key, "'"));
+      }
+      for (const auto& elem : value_node) {
+        if (field->cpp_type() == google::protobuf::FieldDescriptor::CPPTYPE_MESSAGE) {
```
stevenzzzz, line 167: ConvertHeader previously allow scalar value for a key:v pair, now seems it has to be a dict/map to parse correctly.

stevenzzzz (approved):
that's a lot of reflection. :)

In envoy code base, there is a MessageUtil::loadFromYaml method that converts the yaml first to json, then uses standard json-->proto conversion, if this lib wants to be more generic, probably can follow that route.

plugins/test/yaml_proto_converter.cpp
martijneken: That's right. This parser has no custom logic for special fields.

martijneken:
> that's a lot of reflection. :)
> 
> In envoy code base, there is a MessageUtil::loadFromYaml method that converts the yaml first to json, then uses standard json-->proto conversion, if this lib wants to be more generic, probably can follow that route.

Yes, I started there, but the JSON parser doesn't handle binary fields properly: it has no knowledge of types, so it needs to know a priori which proto fields are string vs binary so it can base64 encode the binary ones. That's gross because you end up hard coding the binary fields in the converter. So I thought: why bother with JSON, just look at the fields and types you need to target using reflection. This parser also gives better error messages.

PR #342, opened 2026-05-11
feat: Add a Python Callout implementation for Litellm sample

pweiber:
Adds a litellm_gateway Python sample — a Service Extensions ext_proc callout that turns a Google Cloud external Application Load Balancer into an OpenAI-compatible multi-provider gateway. 

The LB routes each request to the right provider backend (Vertex AI, Anthropic, Groq, OpenRouter) via Internet NEGs; the callout runs LiteLLM in-process to translate OpenAI ↔ provider formats and inject the right auth (ADC for Vertex AI; Secret Manager–backed API keys for the others). 

The callout doesn't proxy traffic — it only mutates the body and headers, and the LB forwards straight to the provider. Routing is header-driven (x-v2-target-provider, derived from the model id)

### DIFF
--- a/callouts/python/extproc/example/litellm_gateway/README.md
+++ b/callouts/python/extproc/example/litellm_gateway/README.md
@@ -1,0 +1,344 @@
+# LiteLLM Gateway
+
+A Service Extensions `ext_proc` callout that turns a Google Cloud external
+Application Load Balancer into an **OpenAI-compatible multi-provider gateway**.
+The load balancer routes each request to the right provider backend (Vertex AI,
+Anthropic, Groq, OpenRouter, …); the callout runs **LiteLLM in-process** to
+translate the OpenAI request/response to/from each provider's native format and
+to inject the right auth.
+
+The callout does **not** proxy traffic — it only mutates headers and the body.
+The load balancer forwards the rewritten request straight to the provider via an
+Internet NEG backend.
+
+## Architecture
+
+```
+                              x-v2-target-provider: anthropic
+  Client  ──────────────────────────────────────────────────────────────┐
+   │  POST /v1/chat/completions   (OpenAI body, model="anthropic/claude…") │
+   ▼                                                                      │
+ ┌────────────────────────────────────────────────────────────────────────┴───┐
+ │                    Global External Application LB                            │
+ │                                                                              │
+ │   ┌──────────────────────────┐        ┌──────────────────────────────────┐   │
+ │   │ URL Map                  │        │  Traffic Extension               │   │
+ │   │  /v1/* + header          │  body  │   Callout (ext_proc, Cloud Run)  │   │
+ │   │   x-v2-target-provider:  ├───────►│   - LiteLLM: provider detect     │   │
+ │   │     anthropic → BE-A     │◄───────┤   - LiteLLM: OpenAI→provider body│   │
+ │   │     groq      → BE-G     │ headers │   - inject auth (ADC / API key)  │   │
+ │   │     openrouter→ BE-O     │  + body │   - rewrite :path / :authority   │   │
+ │   │   /v1/*  (default)→ BE-V │        └──────────────────────────────────┘   │
+ │   │   default → upstream UI  │                                               │
+ │   └────────┬─────────────────┘                                               │
+ │            │  LB forwards to the matched backend                             │
+ │            ▼                                                                 │
+ │   ┌─────────────┐ ┌─────────────┐ ┌─────────────┐ ┌─────────────┐ ┌────────┐ │
+ │   │ BE-V Vertex │ │ BE-A Anthr. │ │ BE-G Groq   │ │ BE-O OpenR. │ │upstream│ │
+ │   │ Internet NEG│ │ Internet NEG│ │ Internet NEG│ │ Internet NEG│ │ UI app │ │
+ │   └──────┬──────┘ └──────┬──────┘ └──────┬──────┘ └──────┬──────┘ └────────┘ │
+ └──────────┼───────────────┼───────────────┼───────────────┼───────────────────┘
+            ▼               ▼               ▼               ▼
+   {region}-aiplatform  api.anthropic.com  api.groq.com   openrouter.ai
+   .googleapis.com
+```
+
+On the response, the LB invokes the callout again; LiteLLM transforms the
+provider's response back to OpenAI format (or, for streaming, parses each SSE
+chunk and re-emits it as an OpenAI `chat.completion.chunk`).
+
+## How It Works — step by step
+
+1. **Client** sends a standard OpenAI request (`POST /v1/chat/completions`,
+   body `{"model": "anthropic/claude-3-5-sonnet-…", "messages": […]}`) plus a
+   routing header `x-v2-target-provider: anthropic`.
+2. **URL Map** evaluates `prefix=/v1/` + `header_matches: x-v2-target-provider`
+   and selects the matching backend service (e.g. the Anthropic Internet NEG).
+   `/v1/*` requests with no provider header fall through to the Vertex AI
+   backend; non-LLM paths go to the upstream UI app.
+3. **Traffic Extension** intercepts the request and streams the headers and
+   body to the callout over gRPC (`REQUEST_HEADERS`, then `REQUEST_BODY`).
+4. **Callout** (`service_callout_example.py`) hands the body to LiteLLM:
+   - `litellm.get_llm_provider(model)` → resolves the provider (`anthropic`).
+   - The provider's `Config` class produces the request body in the provider's
+     native format (`config.transform_request(...)`), the target URL
+     (`config.get_complete_url(...)`), and the auth headers
+     (`config.validate_environment(...)`). For Vertex AI the bearer token is
+     minted from the Cloud Run service identity (ADC); for the others the API
+     key comes from a Secret Manager-backed env var.
+   - The callout returns a `BodyResponse` with the transformed body plus header
+     mutations: `:path`, `:authority`/`host`, the auth header(s),
+     `content-length`, `accept-encoding: identity`, a `user-agent`, and
+     `x-litellm-*` provenance markers. **No `clear_route_cache`** — routing was
+     already decided by the URL map.
+5. **Load balancer** forwards the rewritten request to the chosen backend's
+   Internet NEG, which connects to the provider's API.
+6. **Response phase** — the LB invokes the callout again
+   (`RESPONSE_HEADERS`, `RESPONSE_BODY`). The callout drops the upstream
+   `content-length` (stale after the transform) and, on the body, runs the
+   provider `Config`'s `transform_response(...)` to produce an OpenAI
+   `chat.completion`. For streaming responses it parses each SSE event with the
+   provider's chunk iterator and re-emits an OpenAI `chat.completion.chunk`,
+   appending `data: [DONE]\n\n` at the end.
+7. **Client** receives a standard OpenAI response — same shape regardless of
+   which provider served it.
+
+### The `x-v2-target-provider` header
+
+The client sends `x-v2-target-provider: <provider>` alongside the request — the
+URL map matches on it to pick the backend. The provider is the first segment of
+the LiteLLM model id (`anthropic/claude-…` → `anthropic`), so a one-line
+client-side mapping is all that's required. The sample UI sets it automatically
+from the selected model; OpenAI-style SDK clients can set it via
+`default_headers`. A `/v1/*` request with no provider header falls through to
+the Vertex AI backend.
+
+### Supported endpoints
+
+| Path | Type |
+|------|------|
+| `/v1/chat/completions` | Chat |
+| `/v1/completions` | Text completion |
+| `/v1/embeddings` | Embeddings |
+| `/v1/models` | Model discovery |
+| `/chat/completions` | Chat (alias) |
+| `/completions` | Text completion (alias) |
+| `/embeddings` | Embeddings (alias) |
+
+## Providers
+
+The sample ships with four provider backends. The model id (LiteLLM convention,
+`provider/model`) and the `x-v2-target-provider` header both encode the provider:
+
+| Provider | Example model id | `x-v2-target-provider` | Upstream | Auth |
+|----------|------------------|------------------------|----------|------|
+| Vertex AI | `vertex_ai/gemini-2.5-flash` | `vertex_ai` (or omit — it's the default) | `{region}-aiplatform.googleapis.com` | ADC (Cloud Run service identity, `roles/aiplatform.user`) |
+| Anthropic | `anthropic/claude-3-5-sonnet-20241022` | `anthropic` | `api.anthropic.com` | `ANTHROPIC_API_KEY` env var (Secret Manager) |
+| Groq | `groq/compound-beta` | `groq` | `api.groq.com` | `GROQ_API_KEY` env var (Secret Manager) |
+| OpenRouter | `openrouter/openai/gpt-oss-20b:free` | `openrouter` | `openrouter.ai` | `OPENROUTER_API_KEY` env var (Secret Manager) |
+
+LiteLLM owns the actual translation, so adding a provider that LiteLLM already
+supports needs no callout code change — see below.
+
+### Adding a provider
+
+1. Add the provider's FQDN to `local.third_party_providers` in
+   `deploy/terraform/main.tf` — this creates an Internet NEG + backend service
+   and a URL-map `header_matches` rule for it.
+2. If the provider needs an API key, add a `*_api_key` variable
+   (`deploy/terraform/variables.tf`), wire it into the dynamic `env` block of
+   the callout Cloud Run service, and set it in `terraform.tfvars`. The callout
+   reads it as `<PROVIDER>_API_KEY` automatically.
+3. If LiteLLM's `Config` class for the provider exposes `transform_request`
+   directly (most do — only Vertex Gemini needs the small built-in fallback),
+   nothing else changes.
+
+## Deploy to Google Cloud
+
+> There is intentionally **no local-dev path** for this sample — the ext_proc
+> chain depends on a real GCLB URL map with `header_matches` routing, which an
+> Envoy stand-in doesn't replicate faithfully. Deploy to GCP to exercise it.
+
+### Prerequisites
+
+- [Terraform](https://developer.hashicorp.com/terraform/install) >= 1.0
+- [gcloud CLI](https://cloud.google.com/sdk/docs/install) authenticated with ADC
+- A GCP project with Vertex AI enabled and Gemini accessible in your region
+- API keys for any non-Vertex providers you want to use
+
+#### IAM roles
+
+Scoped for sample/testing deployments — use least-privilege equivalents in
+production.
+
+| Role | Purpose |
+|------|---------|
+| `roles/compute.admin` | Load balancer, backend services, NEGs |
+| `roles/run.admin` | Cloud Run services, revisions, IAM bindings |
+| `roles/networkservices.admin` | Service Extensions traffic extension |
+| `roles/iam.serviceAccountUser` | Bind the Cloud Run service account |
+| `roles/artifactregistry.admin` | Create repos, push/pull images |
+| `roles/secretmanager.admin` | Create secrets for provider API keys |
+| `roles/serviceusage.serviceUsageAdmin` | Enable required GCP APIs |
+| `roles/cloudbuild.builds.editor` | Submit Cloud Build jobs |
+| `roles/storage.admin` | Cloud Build source staging bucket |
+
+The callout runs under the project's default compute service account by default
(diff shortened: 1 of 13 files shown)


callouts/python/extproc/example/litellm_gateway/README.md
```diff
@@ -0,0 +1,344 @@
+# LiteLLM Gateway
+
+A Service Extensions `ext_proc` callout that turns a Google Cloud external
+Application Load Balancer into an **OpenAI-compatible multi-provider gateway**.
+The load balancer routes each request to the right provider backend (Vertex AI,
+Anthropic, Groq, OpenRouter, …); the callout runs **LiteLLM in-process** to
+translate the OpenAI request/response to/from each provider's native format and
+to inject the right auth.
+
+The callout does **not** proxy traffic — it only mutates headers and the body.
```
mpwarres: "The callout implementation does **not** itself proxy traffic, it only mutates headers and the body."

callouts/python/extproc/example/litellm_gateway/README.md
```diff
@@ -0,0 +1,344 @@
...
+The load balancer forwards the rewritten request straight to the provider via an
+Internet NEG backend.
+
+## Architecture
+
+```
+                              x-v2-target-provider: anthropic
+  Client  ──────────────────────────────────────────────────────────────┐
```
mpwarres: There are several lines where the ASCII diagram does not quite line up, because an extra space needs to be added or removed here and there. Can you clean up the ASCII diagram?

callouts/python/extproc/example/litellm_gateway/README.md
```diff
@@ -0,0 +1,344 @@
...
+
+On the response, the LB invokes the callout again; LiteLLM transforms the
+provider's response back to OpenAI format (or, for streaming, parses each SSE
+chunk and re-emits it as an OpenAI `chat.completion.chunk`).
+
+## How It Works — step by step
+
+1. **Client** sends a standard OpenAI request (`POST /v1/chat/completions`,
+   body `{"model": "anthropic/claude-3-5-sonnet-…", "messages": […]}`) plus a
+   routing header `x-v2-target-provider: anthropic`.
```
mpwarres: What does "-v2-" in the header name signify? Can it be removed?

callouts/python/extproc/example/litellm_gateway/README.md
```diff
@@ -0,0 +1,344 @@
...
+   provider `Config`'s `transform_response(...)` to produce an OpenAI
+   `chat.completion`. For streaming responses it parses each SSE event with the
+   provider's chunk iterator and re-emits an OpenAI `chat.completion.chunk`,
+   appending `data: [DONE]\n\n` at the end.
+7. **Client** receives a standard OpenAI response — same shape regardless of
+   which provider served it.
+
+### The `x-v2-target-provider` header
+
+The client sends `x-v2-target-provider: <provider>` alongside the request — the
```
mpwarres: "The client includes `x-v2-target-provider: <provider>` as a request header, and the URL map matches on it to pick the backend."

callouts/python/extproc/example/litellm_gateway/README.md
```diff
@@ -0,0 +1,344 @@
...
+URL map matches on it to pick the backend. The provider is the first segment of
```
mpwarres: If the provider string is always the first segment of the model ID, then I wonder if you could just have a header that gives the entire model ID (`x-model-id` might be a good candidate, seems like that is used elsewhere?), and then use a `prefixMatch` or `regexMatch` in the URL map to match against the header: https://docs.cloud.google.com/load-balancing/docs/url-map-concepts

callouts/python/extproc/example/litellm_gateway/README.md
```diff
@@ -0,0 +1,344 @@
...
+   (`deploy/terraform/variables.tf`), wire it into the dynamic `env` block of
+   the callout Cloud Run service, and set it in `terraform.tfvars`. The callout
+   reads it as `<PROVIDER>_API_KEY` automatically.
+3. If LiteLLM's `Config` class for the provider exposes `transform_request`
+   directly (most do — only Vertex Gemini needs the small built-in fallback),
+   nothing else changes.
+
+## Deploy to Google Cloud
+
+> There is intentionally **no local-dev path** for this sample — the ext_proc
```
mpwarres: Add as the first line of this blockquote:

```
> [!IMPORTANT]
> There is intentionally **no local-dev path** for this sample ...
```

So that it will be formatted as a Markdown callout, per https://docs.github.com/en/get-started/writing-on-github/getting-started-with-writing-and-formatting-on-github/basic-writing-and-formatting-syntax#alerts

callouts/python/extproc/example/litellm_gateway/README.md
```diff
@@ -0,0 +1,344 @@
...
+| Groq | `groq/compound-beta` | `groq` | `api.groq.com` | `GROQ_API_KEY` env var (Secret Manager) |
+| OpenRouter | `openrouter/openai/gpt-oss-20b:free` | `openrouter` | `openrouter.ai` | `OPENROUTER_API_KEY` env var (Secret Manager) |
+
+LiteLLM owns the actual translation, so adding a provider that LiteLLM already
+supports needs no callout code change — see below.
+
+### Adding a provider
+
+1. Add the provider's FQDN to `local.third_party_providers` in
+   `deploy/terraform/main.tf` — this creates an Internet NEG + backend service
```
mpwarres: General comment: there are a lot of em-dashes (i.e. "—") present in this README, which gives it an AI-generated feel. I don't have any problem with AI-generated documentation if is clear and gets the point across, but I'd suggest replacing em-dashes with other punctuation when possible, such as ',', ':', or '.' as appropriate. It makes the documentation feel a little more human.

callouts/python/extproc/example/litellm_gateway/README.md
mpwarres: Adding @arman-rye for opinion on above

callouts/python/extproc/example/litellm_gateway/README.md
pweiber: I believe there is no special meaning to be honest, since it was used in the example by Arman as a custom header I set it like this, but I believe it would work fine without the "-v2-", I can rerun the deploy and tests without it so we can make sure

callouts/python/extproc/example/litellm_gateway/service_callout_example.py
```diff
@@ -0,0 +1,663 @@
...
+
+    def __init__(self, **kwargs) -> None:
+        super().__init__(**kwargs)
+        self.gcp_project = os.getenv("GCP_PROJECT_ID", "")
+        self.gcp_region = os.getenv("GCP_REGION", "us-central1")
+        if not self.gcp_project:
+            logging.warning(
+                "GCP_PROJECT_ID is unset; Vertex AI requests will fail.")
+
+    def process(self, callout, context):
```
mpwarres: If you keep this method override (see later comment), can type annotations be added for the parameters?

callouts/python/extproc/example/litellm_gateway/service_callout_example.py
```diff
@@ -0,0 +1,663 @@
...
+
+        On the response_headers event we tell the LB how to deliver the
+        response body to us: STREAMED for SSE (so we can transform chunks
+        on the fly), BUFFERED for everything else (so the response body
+        arrives whole and we can run LiteLLM's transform_response over it).
+        """
+        resp = super().process(callout, context)
+        if resp.HasField("immediate_response"):
+            return resp
+        if callout.HasField("response_headers"):
```
mpwarres, line 161: Given that this is logic specifically for when processing response headers, why not perform this logic in on_response_headers(), and not override process()?

callouts/python/extproc/example/litellm_gateway/service_callout_example.py
```diff
@@ -0,0 +1,663 @@
...
+        self,
+        body: service_pb2.HttpBody,
+        context: ServicerContext,
+    ) -> service_pb2.BodyResponse | service_pb2.ImmediateResponse | None:
+        state = _state(context)
+        if not state.is_llm:
+            return None
+
+        raw = body.body
+        if not raw:
```
mpwarres, line 214: If this case is unexpected, suggest logging a warning message

callouts/python/extproc/example/litellm_gateway/service_callout_example.py
```diff
@@ -0,0 +1,663 @@
...
+        return self._handle_buffered_chunk(
+            state, body.body or b"", bool(body.end_of_stream))
+
+    # ------------------------------------------------------------- LiteLLM
+
+    def _build_provider_request(
+        self,
+        model: str,
+        req_map: dict,
+    ) -> tuple[str, dict, dict, str, str]:
```
mpwarres: This is a lot of values to put in an unnamed tuple--caller and implementation code could easily get out of sync if a value is added or removed. Suggest using a namedtuple as the return type, and including type annotations for the various fields.

callouts/python/extproc/example/litellm_gateway/service_callout_example.py
```diff
@@ -0,0 +1,663 @@
...
+                StatusCode.InternalServerError)
+
+        state.model = model_name
+        state.provider = provider
+        state.is_streaming = bool(req_map.get("stream"))
+        state.request_body = req_map
+
+        new_body = json.dumps(body_dict).encode("utf-8")
+
+        parsed = urlsplit(api_base)
```
mpwarres: From the implementation of _build_provider_request, it looks like `api_base` holds a URL. If so, then I think it would be more self-explanatory for the var name to end in `_url`. And then `parsed` can also be renamed to `parsed_url`, which is clearer (since many things can be parsed).

callouts/python/extproc/example/litellm_gateway/service_callout_example.py
```diff
@@ -0,0 +1,663 @@
...
+            (HEADER_LITELLM_PROVIDER, provider),
+            (HEADER_LITELLM_MODEL, model_name),
+            (HEADER_LITELLM_STREAMING, "true" if state.is_streaming else "false"),
+        ]
+        # Apply the auth + provider-specific headers LiteLLM computed
+        # (Authorization: Bearer <ADC token>, x-api-key, anthropic-version, …).
+        for k, v in headers_dict.items():
+            if k.lower() in _MANAGED_HEADERS:
+                continue
+            rewrites.append((k.lower(), str(v)))
```
mpwarres: Isn't `v` already a string? Per comment elsewhere, a namedtuple return value from _build_provider_request() with proper type annotations would make the type expectations clear.

callouts/python/extproc/example/litellm_gateway/service_callout_example.py
```diff
@@ -0,0 +1,663 @@
...
+        self,
+        body: service_pb2.HttpBody,
+        context: ServicerContext,
+    ) -> service_pb2.BodyResponse | None:
+        state = _state(context)
+        if not state.is_llm:
+            return None
+        if state.is_streaming:
+            return self._handle_streaming_chunk(
+                state, body.body or b"", bool(body.end_of_stream))
```
mpwarres: Isn't `body.end_of_stream` already a bool?

callouts/python/extproc/example/litellm_gateway/service_callout_example.py
```diff
@@ -0,0 +1,663 @@
...
+        return self._handle_buffered_chunk(
+            state, body.body or b"", bool(body.end_of_stream))
```
mpwarres: Isn't `body.end_of_stream` already a bool?

callouts/python/extproc/example/litellm_gateway/service_callout_example.py
```diff
@@ -0,0 +1,663 @@
...
+        optional_params = {
+            k: v for k, v in req_map.items()
+            if k not in {"model", "messages"}
+        }
+
+        litellm_params: dict = {}
+        is_vertex = provider in ("vertex_ai", "vertex_ai_beta")
+        if is_vertex:
+            # ADC token comes from the Cloud Run service identity at runtime.
+            litellm_params["vertex_project"] = self.gcp_project
```
mpwarres, line 360: Would it make sense to explicitly log an error or throw an exception if these values aren't present, or is that already covered by the `validate_environment()` call?

callouts/python/extproc/example/litellm_gateway/service_callout_example.py
```diff
@@ -0,0 +1,663 @@
...
+                custom_llm_provider="vertex_ai",
+            )
+            headers["Authorization"] = f"Bearer {token}"
+
+        if is_vertex:
+            # VertexGeminiConfig doesn't override get_complete_url; build the
+            # generateContent URL via LiteLLM's internal helper. _get_vertex_url
+            # returns (full_url_with_suffix, suffix_only) — use only the first.
+            from litellm.llms.vertex_ai.common_utils import _get_vertex_url
+            api_base, _ = _get_vertex_url(
```
mpwarres: If `api_base` is a URL, suggest naming it something that ends in `_url`.

callouts/python/extproc/example/litellm_gateway/service_callout_example.py
```diff
@@ -0,0 +1,663 @@
...
+forwards the rewritten request to the provider via an Internet NEG backend.
+
+Routing model: the LB's URL map picks the provider backend from the client's
+`x-v2-target-provider` header — *before* this callout fires. GCP Service
+Extensions don't allow body-based routing on a single LB (route extensions
+can't read the body, traffic extensions can't change the backend), so the
+header is the routing signal. Once the LB has picked the right backend, this
+traffic-extension callout transforms the body and headers in flight.
+
+LiteLLM owns:
```
paul-r-gall, line 30: Is there any quota management here in LiteLLM at all?

callouts/python/extproc/example/litellm_gateway/service_callout_example.py
```diff
@@ -0,0 +1,663 @@
...
+  * provider detection (`litellm.get_llm_provider`)
+  * OpenAI -> provider request body transform
+  * auth (ADC for Vertex AI via Cloud Run service identity, env-var API keys
+    for Anthropic / Groq / OpenRouter / etc.)
+  * provider response -> OpenAI ModelResponse transform
+  * SSE chunk parsing for streaming responses
```
paul-r-gall, line 36: What about observability? Any observability included here?

callouts/python/extproc/example/litellm_gateway/service_callout_example.py
```diff
@@ -0,0 +1,663 @@
...
+        logging.info("Request %s %s", method, path)
+
+        if path not in LLM_ENDPOINTS:
+            return None
+
+        state = _state(context)
+        state.is_llm = True
+
+        resp = service_pb2.ProcessingResponse()
+        resp.request_headers.response.header_mutation.set_headers.append(
```
paul-r-gall, line 192: Do you need to adjust Bearer tokens and API keys in headers?

callouts/python/extproc/example/litellm_gateway/service_callout_example.py
```diff
@@ -0,0 +1,663 @@
...
+            HeaderValueOption(
+                header=HeaderValue(key=HEADER_LITELLM_ROUTED, raw_value=b"true")))
+
+        return resp
+
+    def on_request_body(
```
paul-r-gall, line 198: Please document assumption that request body mode is BUFFERED.

callouts/python/extproc/example/litellm_gateway/service_callout_example.py
paul-r-gall: Ah, I see, you do it below because you're assuming BUFFERED mode.

callouts/python/extproc/example/litellm_gateway/README.md
pweiber: Changed to x-model-id and adjusted the text

callouts/python/extproc/example/litellm_gateway/README.md
pweiber: As we confirmed with Arman I changed it to x-model-id and now it is using prefixMatch

callouts/python/extproc/example/litellm_gateway/README.md
pweiber: As we confirmed with Arman the "-v2-" can be removed and since we decided about the prefixMatch we are now using only the "x-model-id"

callouts/python/extproc/example/litellm_gateway/README.md
pweiber: Done, let me know if it's better

callouts/python/extproc/example/litellm_gateway/README.md
pweiber: Done

callouts/python/extproc/example/litellm_gateway/README.md
pweiber: Done

callouts/python/extproc/example/litellm_gateway/README.md
pweiber: Tried to reword most of the texts, let me know if that's better now

callouts/python/extproc/example/litellm_gateway/service_callout_example.py
pweiber: From what I checked the mode_override lives on the outer ProcessingResponse, not on HeadersResponse, and the base process() just wraps on_response_headers as ProcessingResponse(response_headers=...) with no way to reach mode_override. So I can't set it from on_response_headers without changing the shared base class. I kept the process() override and added the type annotations you asked, if you think it's worth putting up some comments here regarding this let me know

callouts/python/extproc/example/litellm_gateway/service_callout_example.py
pweiber: Done, added a warning

callouts/python/extproc/example/litellm_gateway/service_callout_example.py
pweiber: Done. The ones I kept as api_base are because that's LiteLLM's param name

callouts/python/extproc/example/litellm_gateway/service_callout_example.py
pweiber: validate_environment() doesn't check those. Added a warning when GCP_PROJECT_ID is missing on a Vertex request, plus one at startup. Kept it a warning instead of an exception to match how we handle the empty body case.

callouts/python/extproc/example/litellm_gateway/service_callout_example.py
pweiber: Done. Created the ProviderRequest so it should be better now

callouts/python/extproc/example/litellm_gateway/service_callout_example.py
pweiber: No, there's no quota or rate limiting in the gateway. For this initial sample version we only use LiteLLM for format translation and auth

callouts/python/extproc/example/litellm_gateway/service_callout_example.py
pweiber: Just some x-litellm-* headers (provider, model, streaming) on the forwarded request, plus the routing decision logged to Cloud Run, and LB request logging on so you can see which backend served each request. No metrics or tracing yet since it wasn't in scope for the initial sample version

callouts/python/extproc/example/litellm_gateway/service_callout_example.py
pweiber: Done, let me know if that's better now

callouts/python/extproc/example/litellm_gateway/service_callout_example.py
```diff
@@ -0,0 +1,691 @@
...
+#
+# Anthropic uses /v1/messages directly (not /v1/chat/completions like the
+# OpenAI-compatible providers), so we include the path in the base.
+_PROVIDER_API_BASE = {
+    "anthropic":  "https://api.anthropic.com/v1/messages",
+    "groq":       "https://api.groq.com/openai/v1",
+    "openrouter": "https://openrouter.ai/api/v1",
+}
+
+class ProviderRequest(NamedTuple):
```
mpwarres, line 107: nit: since this is an internal type, its name should probably start with '_'.

callouts/python/extproc/example/litellm_gateway/service_callout_example.py
mpwarres: Thanks for the explanation, that makes sense. I don't think a code comment is needed.

mpwarres (approved):

paul-r-gall (approved):

PR #355, opened 2026-07-30
feat: add opt-in routing, quota, and observability for LiteLLM Sample

pweiber:
Adds 3 optional capabilities to the LiteLLM callout sample. Each one is inert unless configured, so with nothing set the callout behaves exactly as before.                                                                                                                                                                                                                
                                                                                                                                                                                                                                  
**Model routing:** A route extension rewrites the x-model-id header before the URL map evaluates routing, so a request can be steered by model alias, by a tier tag, by a weighted A/B split, or by simple-shuffle across a model group. The rewritten header also becomes authoritative over the request body, so a routing decision changes the model that answers, not just the backend it                                                                                                                                                          
reaches. Enabled by the router_settings section of the gateway config.                                                                                                                                                            
                                                                                                                                                                                                                                  
**Token management:** Virtual keys stored in Memorystore for Redis, enforced in two phases: checked before the upstream call and recorded from the real token usage on the response. Supports token budgets per key and per model, requests                                                                                                                                                     
per minute and tokens per minute limits, model allowlists, key expiry, and warn-only soft budgets, returning 401, 402, 403 and 429 as appropriate.                                                                                                                                                           
Successful responses carry the LiteLLM usage headers. Keys are defined in Terraform and seeded by a Cloud Run Job, or by a seeder script for ad hoc changes. Enabled by REDIS_HOST. 

**Observability:** One OpenTelemetry span per request with GenAI semantic-convention attributes: provider, model, token usage, cost in USD from LiteLLM's bundled price map, a call id that is also returned as a response header, and a truncated hash of the virtual key. Exported over OTLP to any collector, with per-request opt-out. Enabled by OTEL_EXPORTER_OTLP_ENDPOINT.                                                                                                                                                      
                                                                                                                                                                                                                                  
**Regional LB:** A regional Terraform configuration that deploys the gateway with both extensions, Memorystore, and the seed-keys job.

**Docs:** a WALKTHROUGH.md that steps through each capability with the exact curl and log commands and the expected output for the new scenarios.


callouts/python/extproc/example/litellm_gateway/quota.py
```diff
@@ -0,0 +1,303 @@
...
+    if parsed.tzinfo is None:
+        parsed = parsed.replace(tzinfo=timezone.utc)
+    return parsed <= datetime.now(timezone.utc)
+
+
+def check(api_key: str, model: str) -> QuotaDecision:
+    """Request-leg pre-check: unknown/expired key, model allowlist, RPM, TPM,
+    accumulated token budget, and per-model budget (model_max_budget).
+
+    Quota is opt-in per request: a request that presents no virtual key (no
```
leonm1: Quota can't be opt-in per request: if you exceed your quota, you can just remove the API key. Please make it so that quota is either enabled for all or none requests, or provide an opt-in configuration option (i.e. defaulted to off) which allows unauthenticated requests to bypass quota.

callouts/python/extproc/example/litellm_gateway/service_callout_example.py
leonm1: This `optional_params` is passed to LiteLLM's `get_complete_url()` function directly, and `req_map` is parsed from the request body. This allows adversaries to smuggle parameters to get_complete_url and override the behavior of endpoint picking.

Validate all the parameters that get used, and only pass explicitly validated parameters that we plan to support.

callouts/python/extproc/example/litellm_gateway/quota.py
```diff
@@ -0,0 +1,303 @@
...
+
+import hashlib
+import json
+import logging
+import os
+import time
+from datetime import datetime, timezone
+from typing import Any, NamedTuple, Optional
+
+import redis
```
leonm1: This callout server could be fairly high concurrency. We should be using redis.asyncio.Redis rather than sync redis.

leonm1 (changes requested):
Some initial large scoped feedback, I will continue reading through and give more targeted comments as well.

callouts/python/extproc/example/litellm_gateway/service_callout_example.py
```diff
@@ -91,6 +98,25 @@
...
+    "/chat/completions": "chat",
+    "/v1/completions": "text_completion",
+    "/completions": "text_completion",
+    "/v1/embeddings": "embeddings",
+    "/embeddings": "embeddings",
+}
+
+# Quota reject HTTP status codes mapped to Envoy StatusCode values.
+_QUOTA_STATUS = {
+    401: StatusCode.Unauthorized,
```
hillsp: If I understand correctly, `StatusCode.Unauthorized` is [an enum](https://github.com/envoyproxy/envoy/blob/5bec5250cdcf90818533dc2ffe114e89dcfde188/api/envoy/type/v3/http_status.proto) with the value "401". Is the mapping here providing any value? If you are worried that the int must be a valid value in the proto, you could use something like:

```
try:
  StatusCode.Name(http_status_int)
  http_status_valid = True
except ValueError:
  http_status_valid = False
```

or (less readably but apparently faster):

```
http_status_valid = http_status_int in StatusCode.DESCRIPTOR.values_by_number
```

callouts/python/extproc/example/litellm_gateway/gateway_config.py
```diff
@@ -0,0 +1,152 @@
...
+
+
+class QuotaSettings(NamedTuple):
+    fail_open: bool
+
+
+_config: dict[str, Any] = {}
+
+
+def load() -> None:
```
hillsp: Rather than a free function that references a global, I would typically expect callers to use a constructor or factory that returns the config, something like:

```
config = RoutingSettings.load()
```

Is there a reason to prefer the module global? If you need to override this for testing, you can mock out the function.

callouts/python/extproc/example/litellm_gateway/gateway_config.py
```diff
@@ -0,0 +1,152 @@
...
+    if not section:
+        return None
+    tag_rules: dict[tuple[str, str], str] = {}
+    for r in section.get("tag_rules", []):
+        provider_prefix = r.get("provider_prefix") if isinstance(r, dict) \
+            else None
+        tag = r.get("tag") if isinstance(r, dict) else None
+        target = r.get("target") if isinstance(r, dict) else None
+        if not provider_prefix or not tag or not target:
+            logging.warning(
```
hillsp: Rather than logging a warning, what do you think about instead raising for any error you find in the config? Since this parser is run at startup, it seems reasonable to fail to start if the config is bad.

In that world, it's probably a good idea to do this parsing as soon as you load the config (and save the result in a [member] variable), rather than waiting until routing_settings is called.

callouts/python/extproc/example/litellm_gateway/gateway_config.py
```diff
@@ -0,0 +1,152 @@
...
+    A malformed tag_rules entry (missing provider_prefix/tag/target) or
+    weighted_groups member (missing model/weight, or a non-integer weight)
+    is skipped with a warning.
+    """
+    section = _config.get("router_settings")
+    if not section:
+        return None
+    tag_rules: dict[tuple[str, str], str] = {}
+    for r in section.get("tag_rules", []):
+        provider_prefix = r.get("provider_prefix") if isinstance(r, dict) \
```
hillsp: All of this code seems to depend on `isinstance(r, dict)` being true. I think it would make sense to check that up-front and error handle it specifically, then the code below can unconditionally reference `r`. I think that would make the rest of the code easier to read. What do you think?

callouts/python/extproc/example/litellm_gateway/gateway_config.py
```diff
@@ -0,0 +1,152 @@
...
+                "Skipping malformed tag_rules entry (needs "
+                "provider_prefix/tag/target): %r", r)
+            continue
+        tag_rules[(provider_prefix, tag)] = target
+
+    weighted: dict[str, list[tuple[str, int]]] = {}
+    for group, members in section.get("weighted_groups", {}).items():
+        parsed: list[tuple[str, int]] = []
+        for m in members:
+            model = m.get("model") if isinstance(m, dict) else None
```
hillsp: As above on isinstance for `m`.

callouts/python/extproc/example/litellm_gateway/quota.py
```diff
@@ -0,0 +1,303 @@
...
+
+
+_ALLOW = QuotaDecision(True, 0, "")
+
+_client: Optional[Any] = None
+_enabled: bool = False
+_fail_open: bool = True
+
+
+def init_client() -> None:
```
hillsp: As with the config, is there a reason to prefer the module globals here, rather than just returning some class that embeds all these?

callouts/python/extproc/example/litellm_gateway/quota.py
```diff
@@ -0,0 +1,303 @@
...
+    """Connect to Redis once. No-op without REDIS_HOST."""
+    global _client, _enabled, _fail_open
+    if _enabled:
+        return
```
hillsp: Do you need _enabled, or could you just check `_client is not None`?

callouts/python/extproc/example/litellm_gateway/quota.py
```diff
@@ -0,0 +1,303 @@
...
+    unit = window[-1]
+    qty = int(window[:-1])
+    return qty * {"s": 1, "m": 60, "h": 3600, "d": 86400}[unit]
+
+
+def _window_id(window: str) -> int:
+    return int(time.time()) // _window_seconds(window)
+
+
+def _parse_list(raw: str) -> list[str]:
```
hillsp, line 129: Similar to the comment about storing the window as a string, is there a reason that you need to support a comma separated list here? Could you do that parsing in seed_keys and only handle a real array in this code?

callouts/python/extproc/example/litellm_gateway/seed_keys.py
```diff
@@ -0,0 +1,99 @@
...
+        for name in _SCALAR_FIELDS:
+            if name in entry:
+                mapping[name] = str(entry[name])
+        for name in _JSON_FIELDS:
+            if name in entry:
+                mapping[name] = json.dumps(entry[name])
+        if replace:
+            client.delete(f"key:{vk}")
+        client.hset(f"key:{vk}", mapping=mapping)
+        key_hash = hashlib.sha256(vk.encode("utf-8")).hexdigest()[:16]
```
hillsp: What is the purpose of this hash?

callouts/python/extproc/example/litellm_gateway/quota.py
```diff
@@ -0,0 +1,303 @@
...
+        logging.exception("Quota record failed")
+
+
+def usage_headers(api_key: str) -> dict[str, str]:
+    """LiteLLM-style usage response headers for a key; {} when unavailable.
+    """
+    if not _enabled or not api_key:
+        return {}
+    try:
+        cfg = _client.hgetall(f"key:{api_key}")
```
leonm1: This will fetch the config from redis three times per request: once in check(), once in usage_headers(), and once in record() (that'll be for the request headers, response headers, and response body iiuc).

That amplifies the incoming traffic into significantly more per request. We should use a provider with short TTL (1-5s) cache for the config (best) to share it across requests, or at the very least fetch it only once per request and store it in the state of the grpc stream handler (okay).

callouts/python/extproc/example/litellm_gateway/seed_keys.py
```diff
@@ -0,0 +1,99 @@
...
+
+    Raises:
+        ValueError: if an entry is missing the `key` field.
+    """
+    for entry in keys:
+        vk = entry.get("key")
+        if not vk:
+            raise ValueError(f"key entry missing 'key' field: {entry}")
+        mapping: dict[str, str] = {"key": vk}
+        for name in _SCALAR_FIELDS:
```
hillsp: Another option is to structure this so that unknown keys are an error. Something like:

```
for k, v in entry.items():
  if k == "key":
    continue

  if k in _SCALAR_FIELDS:
    mapping[k] = str(v)
  elif k in _JSON_FIELDS:
    mapping[k] = json.dumps(v)
  else:
    raise ValueError(f"Unsupported value key {k} found for redis key {vk}")
```

callouts/python/extproc/example/litellm_gateway/quota.py
```diff
@@ -0,0 +1,303 @@
...
+        if not cfg:
+            return QuotaDecision(False, 401, "unknown key")
+        expires = cfg.get("expires", "")
+        if expires and _expired(expires):
+            return QuotaDecision(False, 401, "key expired")
+        allowed_models = _parse_list(cfg.get("models", ""))
+        if allowed_models and not _model_allowed(model, allowed_models):
+            return QuotaDecision(False, 403, "model not allowed for key")
+        rpm_limit = int(cfg.get("rpm_limit", 0) or 0)
+        if rpm_limit:
```
leonm1: Rather than implememting limits ourselves, we should use a library that implements all of this and is well tested. limits.aio has a redis library that will:
1. Use asyncio for communication with redis, with a single pipelined request rather than multiple round trips
2. Use atomic redis-side accounting for the limits, which is not subject to concurrency races when querying and updating concurrently
3. Has unit and integration tests asserting its correctness

callouts/python/extproc/example/litellm_gateway/service_callout_example.py
```diff
@@ -91,6 +98,25 @@
     "host", ":authority", ":path", "content-length", "content-type",
```
leonm1, line 108: Also `authorization` header?

callouts/python/extproc/example/litellm_gateway/quota.py
```diff
@@ -0,0 +1,303 @@
...
+    _enabled = False
+    _fail_open = True
+
+
+def enabled() -> bool:
+    """Return True if the quota module is active (Redis connected)."""
+    return _enabled
+
+
+def _window_seconds(window: str) -> int:
```
hillsp: This function is necessary because you are storing the window in as a string (`"30d"`) in redis. Is there a reason for that? If you move that parsing into seed_keys.py, you can store the window in redis as a simple int (seconds), reducing runtime risk/complexity.

callouts/python/extproc/example/litellm_gateway/gateway_config.py
```diff
@@ -0,0 +1,152 @@
...
+    """Parsed router_settings, or None to keep the caller's defaults.
+
+    A malformed tag_rules entry (missing provider_prefix/tag/target) or
+    weighted_groups member (missing model/weight, or a non-integer weight)
+    is skipped with a warning.
+    """
+    section = _config.get("router_settings")
+    if not section:
+        return None
+    tag_rules: dict[tuple[str, str], str] = {}
```
leonm1: The second parameter to .get() is a default, but it is not returned in case the yaml section is explicitly set but null. It's safer to explicitly fall back with a falsiness check:

```
# If section.get("tag_rules") returns None, `None or []` evaluates to `[]`
for r in (section.get("tag_rules") or []):
```

callouts/python/extproc/example/litellm_gateway/quota.py
```diff
@@ -0,0 +1,303 @@
...
+    for entry in allowed:
+        if entry == model:
+            return True
+        if entry.endswith("/*") and model.startswith(entry[:-1]):
+            return True
+    return False
+
+
+def _expired(expires: str) -> bool:
+    """True if the ISO-8601 expires timestamp is in the past."""
```
hillsp: I was expecting something a lot simpler, such as:

```
return datetime.fromisoformat(expires) <= datetime.now(timezone.utc)
```

I don't follow exactly what this code is doing with the replaces or why that is necessary. At the very least can you add a code comment explaining why all that parsing is necessary. Better yet, can you update `seed_keys.py` so that the timestamp is in the right format/timezone?

callouts/python/extproc/example/litellm_gateway/routing.py
```diff
@@ -0,0 +1,223 @@
...
+    across requests when clients actually send a distinct id.
+    """
+    model_id = headers.get(ROUTING_HEADER, "")
+    members = _weighted.get(model_id)
+    if not members:
+        return {}
+    total = sum(weight for _, weight in members)
+    if total <= 0:
+        return {}
+    key = headers.get(hash_header) or model_id
```
leonm1: Falling back to the model_id will break split routing. All requests will go to the same backend in the failure case. If there is no header to use for determinism, we should fail safe and use a generated random value for the request.

callouts/python/extproc/example/litellm_gateway/service_callout_example.py
```diff
@@ -395,7 +609,8 @@ def _build_provider_request(
         # Vertex AI's auth lives on the LLM *handler*, not the config: the
         # config's validate_environment doesn't include the ADC bearer token.
         # Call VertexBase directly to mint the token via ADC and inject it.
-        if is_vertex and "Authorization" not in headers and "authorization" not in headers:
+        if (is_vertex and "Authorization" not in headers
```
leonm1: We should normalize the header rather than checking two common capitalisations. Headers are case-insensitive, so we should normalize them to lowercase.

callouts/python/extproc/example/litellm_gateway/service_callout_example.py
```diff
@@ -395,7 +609,8 @@ def _build_provider_request(
...
+                and "authorization" not in headers):
             from litellm.llms.vertex_ai.vertex_llm_base import VertexBase
```
leonm1: All imports at the top please

callouts/python/extproc/example/litellm_gateway/quota.py
hillsp: This treats `rpm_limit: 0` the same as unlimited. Is that what you want? As a user, I'd expect a limit of 0 to mean zero requests. This comment applies in a few places in this function.

callouts/python/extproc/example/litellm_gateway/service_callout_example.py
```diff
@@ -395,7 +609,8 @@ def _build_provider_request(
...
             vb = VertexBase()
```
leonm1: By creating a new VertexBase() for each request, this will require setting up new connections to the Vertex API in each request. The VertexBase object should be a member variable of the server and re-used across requests.

callouts/python/extproc/example/litellm_gateway/quota.py
```diff
@@ -0,0 +1,303 @@
...
+        expires = cfg.get("expires", "")
+        if expires and _expired(expires):
+            return QuotaDecision(False, 401, "key expired")
+        allowed_models = _parse_list(cfg.get("models", ""))
+        if allowed_models and not _model_allowed(model, allowed_models):
+            return QuotaDecision(False, 403, "model not allowed for key")
+        rpm_limit = int(cfg.get("rpm_limit", 0) or 0)
+        if rpm_limit:
+            minute = int(time.time()) // 60
+            rkey = f"rpm:{api_key}:{minute}"
```
hillsp: This mechanism means that the bucket instantly refills to "max" at the start of every minute, which could result in bursty traffic. Is this OK?

The tpm_limit quota below is the same.

callouts/python/extproc/example/litellm_gateway/seed_keys.py
```diff
@@ -0,0 +1,99 @@
...
+            if name in entry:
+                mapping[name] = str(entry[name])
+        for name in _JSON_FIELDS:
+            if name in entry:
+                mapping[name] = json.dumps(entry[name])
+        if replace:
```
leonm1, line 233: The delete + hset need to be in a transaction with pipeline, otherwise there's a race condition if keys are updated while requests are in progress.

callouts/python/extproc/example/litellm_gateway/deploy/terraform-regional/main.tf
```diff
@@ -0,0 +1,896 @@
...
+# tpm_limit, models, model_max_budget, expires, and soft_budget; see the
+# README's "Seeding quota keys" section for the full field table.
+
+resource "google_redis_instance" "quota" {
+  name               = "litellm-gateway-quota"
+  tier               = var.redis_tier
+  memory_size_gb     = 1
+  region             = var.region
+  authorized_network = google_compute_network.vpc.id
+  depends_on         = [google_project_service.apis]
```
leonm1: We should have `auth_enabled = true` and use encryption (`transit_encryption_mode`). It's not safe to have an open-access redis instance with sensitive data, even in a VPC.

callouts/python/extproc/example/litellm_gateway/quota.py
```diff
@@ -0,0 +1,303 @@
...
+    Entries whose value is not itself a dict are dropped, so a malformed
+    entry means "no per-model budget for that model" rather than an
+    exception later in check() or record().
+    """
+    if not raw:
+        return {}
+    try:
+        value = json.loads(raw)
+        if not isinstance(value, dict):
+            return {}
```
hillsp: This seems like it could silently ignore bad config. At least logging seems like it might be useful, but that might be spammy. Is there anything better we can do? 

Similar question on the filter across `isinstance(v, dict)` below.

callouts/python/extproc/example/litellm_gateway/quota.py
```diff
@@ -0,0 +1,303 @@
...
+    it, check() logs a warning but still allows the request.
+
+    Returns status 401 (unknown or expired key), 403 (model not allowed for
+    key), 429 (rpm or tpm exceeded), or 402 (token budget or model budget
+    exceeded) when denied.
+    """
+    if not _enabled or not api_key:
+        return _ALLOW
+    try:
+        cfg = _client.hgetall(f"key:{api_key}")
```
leonm1: Rather than storing the api key in plaintext, this should be a sha-256 hash of the API key to prevent leaking of the API keys.

callouts/python/extproc/example/litellm_gateway/quota.py
```diff
@@ -0,0 +1,303 @@
...
+        if not cfg:
+            return QuotaDecision(False, 401, "unknown key")
```
hillsp: Should this be conditional on `_fail_open`?

callouts/python/extproc/example/litellm_gateway/deploy/terraform-regional/main.tf
```diff
@@ -0,0 +1,896 @@
...
+    google_secret_manager_secret_iam_member.gateway_config_accessor,
+    google_redis_instance.quota,
+  ]
+}
+
+resource "google_cloud_run_v2_service_iam_member" "callout_public_invoker" {
+  name     = google_cloud_run_v2_service.callout.name
+  location = google_cloud_run_v2_service.callout.location
+  role     = "roles/run.invoker"
+  member   = "allUsers"
```
leonm1, line 550: `allUsers` makes this publicly accessible. We should use `default_uri_disabled       = true` and `ingress = "INGRESS_TRAFFIC_INTERNAL_LOAD_BALANCER"` on the cloud run instance to only permit the load balancer access to the cloud run instance, especially since our security posture relies on things that envoy generates i.e. x-request-id

callouts/python/extproc/example/litellm_gateway/deploy/terraform-regional/main.tf
```diff
@@ -0,0 +1,896 @@
...
+  name                  = "litellm-gateway-route-ext"
+  location              = var.region
+  load_balancing_scheme = "EXTERNAL_MANAGED"
+  forwarding_rules = [
+    google_compute_forwarding_rule.forwarding_rule.self_link
+  ]
+  extension_chains {
+    name = "litellm-gateway-route-chain"
+    match_condition {
+      cel_expression = "request.path.startsWith('/v1/')"
```
leonm1: This skips the aliases without the /v1/ prefix

callouts/python/extproc/example/litellm_gateway/service_callout_example.py
```diff
@@ -174,6 +217,39 @@ def on_request_headers(
...
+                )
+            else:
+                logging.info(
+                    "Route extension: no rewrite for %s=%r",
+                    routing.ROUTING_HEADER,
+                    header_map.get(routing.ROUTING_HEADER),
+                )
+            resp = service_pb2.ProcessingResponse()
+            cr = resp.request_headers.response
+            cr.clear_route_cache = True
```
leonm1: We don't need to clear the route cache unless there's a rewrite, right? We should selectively clear it only when needed

callouts/python/extproc/example/litellm_gateway/deploy/terraform-regional/main.tf
```diff
@@ -0,0 +1,896 @@
...
+    sample_rate = 1.0
+  }
+}
+
+# ===================================================================
+# VERTEX AI BACKEND: GLOBAL INTERNET NEG
+# ===================================================================
+#
+# Internet NEGs for provider APIs must be global (INTERNET_FQDN_PORT is a
+# global-only NEG type in the google provider). The regional URL map below
```
leonm1: It says this is a global-only type and says it uses self_link, but I don't see it doing either. It looks like it's registered as a regional NEG and there is no self_link.

That said, it looks like INTERNET_FQDN_PORT is supported by regional LBs: https://docs.cloud.google.com/load-balancing/docs/negs/internet-neg-concepts#internet-neg.

If this works, please fix the comment.

callouts/python/extproc/example/litellm_gateway/telemetry.py
```diff
@@ -0,0 +1,166 @@
...
+    global _tracer, _enabled
+    if _enabled:
+        return
+    endpoint = os.getenv("OTEL_EXPORTER_OTLP_ENDPOINT", "")
+    if not endpoint:
+        return
+    protocol = os.getenv("OTEL_EXPORTER_OTLP_PROTOCOL", "grpc")
+    service_name = os.getenv("OTEL_SERVICE_NAME", "litellm-gateway")
+    if protocol == "grpc":
+        from opentelemetry.exporter.otlp.proto.grpc.trace_exporter import (
```
leonm1: Imports at the top, please

callouts/python/extproc/example/litellm_gateway/service_callout_example.py
```diff
@@ -174,6 +217,39 @@ def on_request_headers(
...
+                    HeaderValueOption(
+                        header=HeaderValue(
+                            key=k, raw_value=v.encode("utf-8")),
+                        append_action=(
+                            HeaderValueOption
+                            .OVERWRITE_IF_EXISTS_OR_ADD)))
+            return resp
+
         path, method = "", ""
         for h in headers.headers.headers:
```
leonm1: We loop over headers.headers.headers three times: once on line 223, once here, and once on line 267. Can we do it just once on 223, then use dict getters to get path, method, and authorization headers?

callouts/python/extproc/example/litellm_gateway/gateway_config.py
```diff
@@ -0,0 +1,152 @@
...
+
+    A bad path or a YAML syntax error is logged as a warning and leaves the
+    config empty (routing disabled, other defaults apply).
+    """
+    path = os.getenv("GATEWAY_CONFIG", "")
+    if not path:
+        return
+    try:
+        with open(path, "r", encoding="utf-8") as f:
+            _load_dict(yaml.safe_load(f) or {})
```
leonm1: This silently falls back to default config. We probably should not do that -- failure to load the specified config should cause the server to log a clear error error and exit

callouts/python/extproc/example/litellm_gateway/gateway_config.py
```diff
@@ -0,0 +1,152 @@
...
+            for k, v in section.get("shuffle_groups", {}).items()
+        },
+    )
+
+
+def telemetry_settings() -> TelemetrySettings:
+    section = _config.get("litellm_settings") or {}
+    return TelemetrySettings(
+        redact_user_api_key_info=bool(
+            section.get("redact_user_api_key_info", False)))
```
leonm1: This should default to True

callouts/python/extproc/example/litellm_gateway/gateway_config.py
```diff
@@ -0,0 +1,152 @@
...
+
+
+def quota_settings() -> QuotaSettings:
+    section = _config.get("general_settings") or {}
+    return QuotaSettings(fail_open=bool(section.get("quota_fail_open", True)))
```
leonm1: LLMs are expensive, we should probably fail closed

callouts/python/extproc/example/litellm_gateway/telemetry.py
```diff
@@ -0,0 +1,166 @@
...
+    operation: str = "chat",
+    api_key: str = "",
+    call_id: str = "",
+) -> None:
+    """Set GenAI attributes and end the span. No-op on a None span."""
+    if span is None:
+        return
+    span.set_attribute("gen_ai.operation.name", operation)
+    span.set_attribute("gen_ai.system", provider or "")
+    span.set_attribute("gen_ai.request.model", model or "")
```
leonm1: Should we also log original model from x-request-id?

callouts/python/extproc/example/litellm_gateway/telemetry.py
```diff
@@ -0,0 +1,166 @@
...
+    if completion_tokens is not None:
+        span.set_attribute("gen_ai.usage.output_tokens", completion_tokens)
+    if total_tokens is not None:
+        span.set_attribute("gen_ai.usage.total_tokens", total_tokens)
+    cost = _completion_cost(
+        provider, model, prompt_tokens, completion_tokens)
+    if cost is not None:
+        span.set_attribute("gen_ai.cost.total_cost", cost)
+    if call_id:
+        span.set_attribute("litellm.call_id", call_id)
```
leonm1: Should we also log the strategy that caused the model to be chosen?

callouts/python/extproc/example/litellm_gateway/README.md
```diff
@@ -306,33 +315,354 @@ gcloud storage rm -r gs://YOUR_PROJECT_ID_cloudbuild/
...
+| `x-litellm-call-id` | Per-request UUID, the same value attached to the trace span as `litellm.call_id`. |
+
+As with the LiteLLM Proxy's own usage headers, these values reflect spend
+recorded so far: the response that completes the in-flight request carries
+headers that do not yet include that request's own usage, since spend is
+recorded after the response body is read.
+
+### Telemetry (OpenTelemetry)
+
+Each LLM request emits one span (`llm.request`) with GenAI
```
leonm1: The opentelemetry spec says model requests should be logged with a specific span name:

https://github.com/open-telemetry/semantic-conventions-genai/blob/main/docs/gen-ai/gen-ai-spans.md#spans

`{gen_ai.operation.name} {gen_ai.request.model}`

callouts/python/extproc/example/litellm_gateway/README.md
```diff
@@ -306,33 +315,354 @@ gcloud storage rm -r gs://YOUR_PROJECT_ID_cloudbuild/
...
+| `x-ratelimit-remaining-tokens` | `token_budget` minus spend so far. Present only when the key has a `token_budget`. |
+| `x-ratelimit-remaining-requests` | `rpm_limit` minus requests so far this minute. Present only when the key has an `rpm_limit`. |
+| `x-litellm-call-id` | Per-request UUID, the same value attached to the trace span as `litellm.call_id`. |
+
+As with the LiteLLM Proxy's own usage headers, these values reflect spend
+recorded so far: the response that completes the in-flight request carries
+headers that do not yet include that request's own usage, since spend is
+recorded after the response body is read.
+
+### Telemetry (OpenTelemetry)
```
leonm1, line 509: This exposes spans, but we should also expose metrics: https://github.com/open-telemetry/semantic-conventions-genai/blob/main/docs/gen-ai/gen-ai-metrics.md

callouts/python/extproc/example/litellm_gateway/quota.py
```diff
@@ -0,0 +1,303 @@
...
+    """Response-leg update: add token usage to spend and TPM counters, plus
+    the per-model spend counter when the key has a model_max_budget entry
+    for this model."""
+    if not _enabled or not api_key or total_tokens <= 0:
+        return
+    try:
+        cfg = _client.hgetall(f"key:{api_key}")
+        window = cfg.get("budget_duration", "30d") if cfg else "30d"
+        skey = f"spend:{api_key}:{_window_id(window)}"
+        if _client.incrby(skey, total_tokens) == total_tokens:
```
hillsp: Should this be `>=` instead of `==`? This applies here and 2 more places below too.

callouts/python/extproc/example/litellm_gateway/quota.py
hillsp: You call this `hgetall` 3 times per request. What are the performance implications of that? Should we cache the result so it's only once?

callouts/python/extproc/example/litellm_gateway/service_callout_example.py
```diff
@@ -174,6 +217,39 @@ def on_request_headers(
         headers: service_pb2.HttpHeaders,
         context: ServicerContext,
     ) -> service_pb2.ProcessingResponse | None:
+        state = _state(context)
+        if state.route_mode:
+            header_map = {
```
hillsp: I think it would be good to move this route mode handling to its own function. WDYT?

callouts/python/extproc/example/litellm_gateway/service_callout_example.py
```diff
@@ -220,21 +315,56 @@ def on_request_body(
             req_map = json.loads(raw)
         except json.JSONDecodeError as e:
             logging.warning("Invalid JSON body: %s", e)
-            return callout_tools.header_immediate_response(StatusCode.BadRequest)
+            self._end_span(state, 400)
```
hillsp: WDYT about a single function that both updates the telemetry and calls `header_immediate_response`? This seems especially useful now that we know that StatusCode.BadRequest == 400.

callouts/python/extproc/example/litellm_gateway/service_callout_example.py
```diff
@@ -588,19 +844,25 @@ def _stream_iterator(self, state: _StreamState):
             return state.stream_iterator
         state.stream_iterator_resolved = True
         if state.provider in ("vertex_ai", "vertex_ai_beta"):
-            from litellm.llms.vertex_ai.gemini.vertex_and_google_ai_studio_gemini import (
-                ModelResponseIterator,
+            from litellm.llms.vertex_ai.gemini import (
```
hillsp: There is a performance penalty to importing a python package inline, rather than once at the top of the file, and this is a hot function. Is that a reason you decided to do it this way?

callouts/python/extproc/example/litellm_gateway/quota.py
pweiber: Thanks for all the review, for the async redis I believe we would need an await point in the callout, and our current sdk is still sync. I created a draft to move it to asyncio in #315  while we were load testing it some months ago. I think this affects limits.aio too. What do you think?

callouts/python/extproc/example/litellm_gateway/telemetry.py
```diff
@@ -0,0 +1,166 @@
...
+from extproc.example.litellm_gateway import gateway_config
+
+_tracer: Optional[trace.Tracer] = None
+_enabled: bool = False
+
+
+def init_tracer() -> None:
+    """Initialize the global tracer once. No-op without an OTLP endpoint."""
+    global _tracer, _enabled
+    if _enabled:
```
hillsp: Do we need `_enabled`, or can we just check if `_tracer is None`?

callouts/python/extproc/example/litellm_gateway/telemetry.py
```diff
@@ -0,0 +1,166 @@
...
+        )
+    else:
+        from opentelemetry.exporter.otlp.proto.http.trace_exporter import (
+            OTLPSpanExporter,
+        )
+    provider = TracerProvider(
+        resource=Resource.create({"service.name": service_name}))
+    provider.add_span_processor(BatchSpanProcessor(OTLPSpanExporter()))
+    _install_provider(provider)
+    atexit.register(provider.shutdown)
```
hillsp: Do you need this? The internet seems to think that TracerProvider will do this for you.

callouts/python/extproc/example/litellm_gateway/quota.py
leonm1: I will review #315 in parallel, it looks fairly small in comparison. We can implement the handler here with asyncio and, worst case, run the async handlers from a sync method with `asyncio.run()`. It'll probably be slower in the short term (waiting for async from sync provides no performance benefit afaict), but once we can land #315 we can make it async native, which will greatly increase performance.

callouts/python/extproc/example/litellm_gateway/quota.py
pweiber: Done, let me know if that's fine now

callouts/python/extproc/example/litellm_gateway/service_callout_example.py
pweiber: Done, let me know if that's fine now

callouts/python/extproc/example/litellm_gateway/quota.py
pweiber: Done, I tried **asyncio.run()** per call and it does not survive a second request. It closes its loop on return while the Redis pool outlives it, so the next call gets a connection bound to a dead loop: 
`RuntimeError: Event loop is closed`
then got 
`Future attached to a different loop`
Running the coroutines on one long-lived loop instead fixed it with no other change. That loop is ~20 lines and I believe it can be deleted the moment #315 lands.

callouts/python/extproc/example/litellm_gateway/quota.py
pweiber: Done, let me know if that's fine now

callouts/python/extproc/example/litellm_gateway/quota.py
pweiber: Adopted for `rpm_limit`: `limits.aio` with `MovingWindowRateLimiter` over the same Redis. But for `tpm_limit` did not fit and stays a counter. `limits` declines to record a hit whose cost exceeds the limit, and a single response routinely costs more than a whole minute's token allowance, so the window never filled and the limit never fired: `hit(cost=30)` against a limit of 10 returns False and leaves `remaining: 10`. From the test I did, tpm went [200, 200, 200] on the library and [200, 429, 429] on the counter. Maybe after the #315 we can try again to see if it works properly or if it was a configuration issue from my side

callouts/python/extproc/example/litellm_gateway/service_callout_example.py
pweiber: I believe it could not go in that set. `_MANAGED_HEADERS (:107)` is a skip-list over LiteLLM's output headers, since the callout owns` :path`,` :authority` and the content headers itself, adding `authorization` there would skip LiteLLM's own `Authorization: Bearer <ADC token>` and send every Vertex request out unauthenticated. It is handled at `:525` instead: when LiteLLM emits an `authorization` the rewrite overwrites the caller's value (`OVERWRITE_IF_EXISTS_OR_ADD`), and when the provider authenticates some other way and emits none (Anthropic's `x-api-key`), `authorization` is added to `remove_headers`. Either way the caller's virtual key stops at the gateway.

callouts/python/extproc/example/litellm_gateway/gateway_config.py
pweiber: Done, let me know if that's fine now

callouts/python/extproc/example/litellm_gateway/routing.py
pweiber: Done, let me know if that's fine now

callouts/python/extproc/example/litellm_gateway/service_callout_example.py
pweiber: Done, let me know if that's fine now

callouts/python/extproc/example/litellm_gateway/service_callout_example.py
pweiber: Done, let me know if that's fine now

callouts/python/extproc/example/litellm_gateway/seed_keys.py
pweiber: Done, let me know if that's fine now

callouts/python/extproc/example/litellm_gateway/deploy/terraform-regional/main.tf
pweiber: Done, let me know if that's fine now

callouts/python/extproc/example/litellm_gateway/quota.py
pweiber: Done, let me know if that's fine now

callouts/python/extproc/example/litellm_gateway/deploy/terraform-regional/main.tf
pweiber: Ingress and `default_uri_disabled` are in. On `allUsers`: I removed the binding on a live deployment and every request returned 500, with Cloud Run logging `The request was not authenticated. Either allow unauthenticated invocations or set the proper Authorization header. Empty Authorization header value`. 

Service Extensions calls ext_proc without an identity token, so there is nothing for IAM to match. I also granted `roles/run.invoker` to the compute service account instead, still 500 on every request, and there is no `gcp-sa-networkservices` service agent in the project to grant it to. Restoring the binding recovered immediately. 

I was taking a look at the callouts docs and found: "the Cloud Run service must be configured to accept HTTP/2 traffic and must allow unauthenticated access" ([Callouts overview](https://cloud.google.com/service-extensions/docs/callouts-overview), under "Supported backends for user-managed callout backend services"). And the Cloud Run ingress docs describe this exact pairing: "For a layered approach to managing access, use both network ingress settings and IAM authentication" ([Restricting ingress](https://cloud.google.com/run/docs/securing/ingress)). With INGRESS_TRAFFIC_INTERNAL_LOAD_BALANCER there is no public path to reach it. Is there another config I should be applying?

callouts/python/extproc/example/litellm_gateway/deploy/terraform-regional/main.tf
pweiber: Did some changes, let me know if that's fine now

callouts/python/extproc/example/litellm_gateway/service_callout_example.py
pweiber: Correct, let me know if that's fine now

callouts/python/extproc/example/litellm_gateway/deploy/terraform-regional/main.tf
pweiber: The comments should be fixed now, let me know if you find anything else

callouts/python/extproc/example/litellm_gateway/service_callout_example.py
pweiber: Done, let me know if that's fine now

callouts/python/extproc/example/litellm_gateway/gateway_config.py
pweiber: Done, let me know if that's fine now

callouts/python/extproc/example/litellm_gateway/gateway_config.py
pweiber: Done, let me know if that's fine now

callouts/python/extproc/example/litellm_gateway/gateway_config.py
pweiber: Done, let me know if that's fine now

callouts/python/extproc/example/litellm_gateway/telemetry.py
pweiber: Added as `litellm.request.original_model`, under `litellm.`

callouts/python/extproc/example/litellm_gateway/telemetry.py
pweiber: Added as `litellm.routing_strategy`

callouts/python/extproc/example/litellm_gateway/README.md
pweiber: Done, let me know if that's fine now

callouts/python/extproc/example/litellm_gateway/README.md
pweiber: Done, let me know if that's fine now

callouts/python/extproc/example/litellm_gateway/service_callout_example.py
pweiber: No value, the enum values are the HTTP codes, so it was an identity map. Dropped, and the status is passed straight through, with a values_by_number guard so an unmapped code degrades to 500 rather than raising inside the proto.

callouts/python/extproc/example/litellm_gateway/gateway_config.py
pweiber: Done, `GatewayConfig.load()` and `.from_dict()` are classmethods that return a value, and the module global is gone. Same shape applied to quota, telemetry and routing.

callouts/python/extproc/example/litellm_gateway/gateway_config.py
pweiber: Agreed, the whole file is parsed in `from_dict` at startup and anything unusable raises

callouts/python/extproc/example/litellm_gateway/gateway_config.py
pweiber: Done, `_section()` does the check once and every parser entry point goes through it, so a non-mapping section fails with the section name rather than an `AttributeError` deeper in.

callouts/python/extproc/example/litellm_gateway/gateway_config.py
pweiber: Same fix, a non-mapping member raises with the group name and the offending entry, instead of being assumed to be a mapping.

callouts/python/extproc/example/litellm_gateway/quota.py
pweiber: No good reason to be honest. `quota`, `gateway_config`, `telemetry` and `routing` are classes now, built in the callout's `__init__`.

callouts/python/extproc/example/litellm_gateway/quota.py
pweiber: Gone, a `Quota` with no client is now a working no-op, so `enabled()` is just `self._client is not None` and nothing branches on it.

callouts/python/extproc/example/litellm_gateway/quota.py
pweiber: No reason,  the seeder now takes either shape from YAML, validates it and writes a JSON array; `_parse_list` is `json.loads(raw) if raw else []`, and a bad value fails the seed run with the field named.

callouts/python/extproc/example/litellm_gateway/seed_keys.py
pweiber: It is the Redis key name now, virtual keys are credentials, so a dump of the datastore shows usage but no keys. The log line uses a truncated copy.

callouts/python/extproc/example/litellm_gateway/seed_keys.py
pweiber: An unknown field is now an error at seed time, named alongside the supported list, both for the key itself and for `model_max_budget` entries.

callouts/python/extproc/example/litellm_gateway/quota.py
pweiber: Done, rather than relocating the parser I deleted it, so keys.yaml is seconds too. I did start with `30d` since LiteLLM accepts that in /key/generate, but one representation means there is no unit parser to keep correct, and a bad value now fails the seed run (`budget_duration: expected an integer, got '30d'`) instead of at request time. The tradeoff is that a keys file written for LiteLLM errors instead of converting

callouts/python/extproc/example/litellm_gateway/quota.py
pweiber: Done, it is that one line now. seed_keys.py pins the timestamp to UTC on write.

callouts/python/extproc/example/litellm_gateway/quota.py
pweiber: Done, let me know if that's fine now

callouts/python/extproc/example/litellm_gateway/quota.py
pweiber: Fixed for `rpm_limit`, which now uses a moving window via `limits` (leonm1's suggestion above), so straddling a minute boundary no longer buys a second allowance. `tpm_limit` keeps a fixed window, for the reason in that thread.

callouts/python/extproc/example/litellm_gateway/quota.py
pweiber: Validated in `seed_keys.py` now, so a malformed budget fails the seed run with the offending field named. The request path should no longer have a bad-config case

callouts/python/extproc/example/litellm_gateway/quota.py
pweiber: I re-checked, and it should not. `fail_open` covers the `except` path, where Redis is unreachable and we have no answer. An unknown key is an answer, from a healthy backend. Making it conditional would mean any unrecognized string becomes a valid key whenever `fail_open` is on

callouts/python/extproc/example/litellm_gateway/quota.py
pweiber: Agreed, applied to all three, let me know if that's fine now

callouts/python/extproc/example/litellm_gateway/quota.py
pweiber: Agreed, did the changes, let me know if that's fine now

callouts/python/extproc/example/litellm_gateway/service_callout_example.py
pweiber: Agreed, it is `_route_extension_response()` now.

callouts/python/extproc/example/litellm_gateway/service_callout_example.py
pweiber: Agreed, it is `_reject(state, status)` now

callouts/python/extproc/example/litellm_gateway/service_callout_example.py
pweiber: No good reason, it was a misconfiguration from my IDE, should be fine now

callouts/python/extproc/example/litellm_gateway/telemetry.py
pweiber: Gone, same as in quota. `enabled()` is `self._tracer is not None`, and a Telemetry with no tracer is a working no-op, so nothing branches on it.

callouts/python/extproc/example/litellm_gateway/telemetry.py
pweiber: Just checked, the `TracerProvider.__init__` defaults `shutdown_on_exit=True` and registers it already. Removed the duplicate.

callouts/python/extproc/example/litellm_gateway/quota.py
hillsp: Thanks for the change here and in `_parse_model_budgets`. I notice that you removed the try/catch for JSONDecodeError. Are you confident that we don't need that any more, or should these return empty in that case?

callouts/python/extproc/example/litellm_gateway/quota.py
hillsp: Appreciate you moving this out of the request parsing. If it turns out that we need to consume a LiteLLM keys file, we could easily add that parsing into seed_keys.py in the future, without breaking anything. Thanks!

callouts/python/extproc/example/litellm_gateway/quota.py
hillsp: Thanks for those changes - that should have improved the performance for sure.

Looking at the code you have now, I think it would be a nice cleanup to wrap the API key, model and config into a class, something like this (much simplified):

```
class QuotaHandler:
  @classmethod
  def create(cls, async_bridge, quota, api_key, model):
    cfg = async_bridge.run(quota.key_config(state.api_key))
    return QuotaHandler(async_runner, quota, api_key, model, cfg)

  def check(self):
    try:
      return self.async_bridge.run(self.quota.check(self.api_key, self.model, self.cfg))
    except Exception:
      return self.quota.unavailable()   

if self.quota.enabled:
  state.quota_handler = QuotaHandler.create(self._async, self.quota, state.api_key, model)
  decision = state.quota_handler.check()
else:
  sate.quota_handler = None

[...]

if state.quota_handler:
  self.quota_handler.record()
```

That's a totally optional suggestion - what you have solves the performance problem.

hillsp (commented):
Thanks for the improvements! I would have resolved a lot of the comment threads but I it turns out that I don't have permission. Please resolve anything that I didn't explicitly comment on.

